PR #39, opened 2014-12-21
CON-188. Add LIKE and NOT_LIKE operators as aliases to REGEX and NOT_REGEX

remiemalik:
I have added the two operators. Should i unit test this, or is this case a bit to small?
Let me know.

### DIFF
--- a/concourse-integration-tests/src/test/java/org/cinchapi/concourse/OperatorAliasTest.java
+++ b/concourse-integration-tests/src/test/java/org/cinchapi/concourse/OperatorAliasTest.java
@@ -1,0 +1,42 @@
+package org.cinchapi.concourse;
+
+import org.cinchapi.concourse.thrift.Operator;
+import org.cinchapi.concourse.util.TestData;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Unit tests the following alias operators: LIKE, NOT_LIKE and LINK_TO.
+ * 
+ * @author remie
+ *
+ */
+public class OperatorAliasTest extends ConcourseIntegrationTest {
+
+	@Test
+	public void testLikeAlias() {
+		String key = TestData.getString();
+		int record = TestData.getScaleCount();
+		client.add(key, 1, record);
+		Assert.assertEquals(client.find(key, Operator.LIKE, record),
+				client.find(key, Operator.REGEX, record));
+	}
+
+	@Test
+	public void testNotLikeAlias() {
+		String key = TestData.getString();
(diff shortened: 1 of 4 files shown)


remiemalik:
I have added the switch method (089390f). I see that it is much cleaner.

jtnelson:
Awesome. Last thing to do is add some unit tests to verify the functionality. 

You should edit StoresTest#testNormalizeOperator to include a switch statement similar to the one you added in Stores.java.

You should also add a class to the concourse-integration-tests project. Call it something like `LikeNotLikeAliasTest` and add some test methods that ensure that the results from using LIKE is the same as the results of using REGEX and the results of using NOT_LIKE are the same as using NOT_REGEX.

You can look at any of the test classes in the org.cinchapi.concourse package for inspiration. Your test class will need to extend `ConcourseIntegrationTest`. That will give you access to a `client` variable in each of your test methods which you can use to add data to Concourse and then perform queries. Let me know if you have any trouble with this and I'll gladly help!

remiemalik:
Thanks for the help. It was very usefull  looking at the existing examples.
I have added the unit tests. Figured to also unit test the LINK_TO alias. 

Let me know what you think ;)

jtnelson:
Couple more changes:

1. Can you switch to using the Cinchapi formatter for the concourse projects so your spacing, etc is consistent with the rest of the codebase? You can find the formatter at https://raw.githubusercontent.com/cinchapi/dev-tools/master/CinchapiFormatter.xml and you can find instructions for installing project specific formatters (so it doesn't mess up your global eclipse settings) at http://stackoverflow.com/a/16045824.

2. Your unit tests are passing, but they are doing so trivially. Try to construct scenarios in the cases that are designed to actually test what the operators are looking for. For example, the LIKE operator should return the same thing as the REGEX operator. And, the REGEX operator should return records that match a regular expression and exclude records that do not match the regular expression.

So, you could write a test case like:
```
client.add("foo", "barfoobar", 1);
client.add("foo", "barbazbar", 2);
client.add("foo", "bazfoobar", 3);
client.add("foo", "foobarfoo", 4);
client.add("foo", "bazbarbaz", 5);

Assert.assertEquals(client.find("foo", Operator.LIKE, "%foo%"), client.find("foo", Operator.REGEX, "%foo%");
```

That way your tests will actually have results sets with meaningful results.

remiemalik:
Yes, ofcourse. I have added the formatter for each concourse project.

I did not realize that it would be better to add multiple records. Thanks for pointing that out.
Still learning step by step. Hope these commits are allright.

PR #47, opened 2015-01-23
CON: 199 Added Unsafe methods for each read operations

dubex:


jtnelson:
Can you make the "safe" methods call into the unsafe methods. Similar to how the add method calls addUnsafe after it grabs the locks. Also add some javadoc comments to the unsafe methods.

dubex:
Why we are calling unsafe methods from safe methods, is it something
related to clean code?

jtnelson:
Exactly! If we need to add functionality to the routines in the future,
we'll only have to add that code in one place instead of two. The way
add/remove works are good examples. In these cases, we have to notify
version change listeners about the update. That code is only placed in the
addUnsafe method and the add method just calls into that after is has done
the necessary locking to make it "safe"

PR #52, opened 2015-02-01
CON: 199 Added Unsafe methods for each read operations

dubex:


jtnelson:
1. Can you apply the Cinchapi Formatter (https://raw.githubusercontent.com/cinchapi/dev-tools/master/CinchapiFormatter.xml) so your spacing is uniform with the rest of the class?

concourse-server/src/main/java/org/cinchapi/concourse/server/storage/Engine.java
```diff
@@ -388,13 +388,27 @@ public void addVersionChangeListener(Token token,
...
+    /**
+     * This method returns a log of revisions in {@code record} as
+     * a Map WITHOUT grabbing any locks. This method is ONLY appropriate
+     * to call from the methods of {@link #AtomicOperation} class
+     * because in this case intermediate lock is not required. 
+     * 
+     * @param record
+     * @return {@code Map}
+     */
+    public Map<Long, String> auditUnsafe(long record) {
```
jtnelson: Can you make all the "unsafe" methods you added have package-private (default) protection instead of public.

jtnelson:
I just realized this, but all the "unsafe" methods still need to grab the transportLock

jtnelson:
I just thought about this, and I'm worried that the "safe" methods grabbing the read lock before calling into the unsafe methods which grab the transportLock might cause some potential deadlocks. Therefore, I think your first approach where the safe methods DONT call into the unsafe methods is ideal. So things would look something like:

```
   @Override
    public Map<TObject, Set<Long>> browse(String key) {
        transportLock.readLock().lock();
        Lock read = lockService.getReadLock(key);
        read.lock();
        try {
            return super.browse(key);
        }
        finally {
            read.unlock();
            transportLock.readLock().lock();
       }
    }

    Map<TObject, Set<Long>> browseUnsafe(String key) {
       transportLock.readLock().lock();
       try{
           super.browse(key);
       }
       finally {
           transportLock.readLock().unlock();
       }
    }
```

PR #56, opened 2015-02-03
Added support for specifying Operators as symbols in CriteriaBuilder

jamalmoir:


concourse/src/main/java/org/cinchapi/concourse/lang/KeyState.java
```diff
@@ -53,4 +54,8 @@ public OperatorState operator(Operator operator) {
         return new OperatorState(criteria);
     }
```
jtnelson: Can you add some javadoc to this method?

PR #59, opened 2015-02-10
Concourse server uses newest version of Concourse-cli, 1.1.3

harymitchell:
* Release required to use newest version of concourse-import.


harymitchell:
Didnt realize I was so behind.  I'll have to merge.

PR #60, opened 2015-02-11
CON-217 Fixed transaction memory leak bug

dubex:


concourse-server/src/main/java/org/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -824,11 +826,13 @@ public void set0(String key, TObject value, long record, AccessToken creds,
...
             throws TException {
         checkAccess(creds, null);
-        TransactionToken token = new TransactionToken(creds, Time.now());
+        Transaction existing = transactions.get(creds);
+        if(existing != null) {
+            existing.abort();
+        }
         Transaction transaction = getEngine(env).startTransaction();
-        transactions.put(token, transaction);
-        Logger.info("Started Transaction {}", transaction);
```
jtnelson: Can you add this log message back

dubex:
added log message back.

PR #61, opened 2015-02-24
CON-44 : Add diff methods to public API

dubex:


jtnelson:
can you add some unit tests :)

PR #64, opened 2015-03-13
public browse()added in Concourse.java

vijaykakadiya:


concourse/src/main/java/org/cinchapi/concourse/Concourse.java
```diff
@@ -1633,6 +1640,11 @@ public Object apply(TObject input) {
         public Map<Object, Set<Long>> browse(String key) {
             return browse(key, now);
         }
+        
+        @Override
+        public Set<Long> browse() {
+            return browse();
```
jtnelson: this is going to StackOverflow from infinite recursion. Look at the implementation for `public Map<String, Set<Object>> browse(final long record, final Timestamp timestamp)` in Concourse#Client to see how you need to use execute to actually make the RPC call.

jtnelson:
I added this when I re-wrote the Thrift API :)

PR #68, opened 2015-03-24
Defined Endpoint for audit and chronologize method

dubex:


jtnelson:
Can you rebase against the latest changes to IndexRouter? The way to parse timestamps has changed. You should use `Timestamp#parse` now instead of `Long#tryParse`

PR #70, opened 2015-04-07
Defined thrift API and Endpoint for diff method

dubex:


jtnelson:
Can you generate the thrift API with 

`./utils/compile-thrift-all.sh`

concourse-driver-java/src/main/java/org/cinchapi/concourse/Concourse.java
```diff
@@ -504,8 +504,8 @@ public final void close() {
      * @param end
      * @return the changes made to the record within the range
      */
-    public abstract Map<String, Map<Diff, Set<Object>>> diff(long record,
-            Timestamp start, Timestamp end);
+    public abstract <T> Map<String, Map<Diff, Set<T>>> diff(final long record,
```
jtnelson: Don't define the parameters as final in the abstract method signature because that is an implementation detail that the abstract class doesn't need to be aware of

concourse-server/src/main/java/org/cinchapi/common/util/KeyRecordArgs.java
```diff
@@ -0,0 +1,77 @@
+package org.cinchapi.common.util;
```
jtnelson: I don't really think this needs to be defined as a top level class since its very specific to the http code . But, if you think it can be used in other places down the line, you should define it in a class like org.`cinchapi.concourse.server.util` since it won't be used outside of Concourse. Also, if you think it should be a more generic class, probably a good idea to change the name to something like "DataArgs"

concourse-server/src/main/java/org/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -87,12 +88,15 @@
 import org.cinchapi.concourse.util.DataServices;
 import org.cinchapi.concourse.util.Environments;
 import org.cinchapi.concourse.util.Logger;
+import org.cinchapi.concourse.util.PrettyLinkedHashMap;
```
jtnelson: Shouldn't use Pretty maps on the server because the client will still need to transform the data, so we don't want to waste the CPU cycles on the server to format the data in the map. Furthermore, Pretty maps are only meant to help with console display, so the server shouldn't really be concerned with that. You should use a LinkedHashMap instead.

concourse-server/src/main/java/org/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -1147,6 +1152,118 @@ public boolean commit(AccessToken creds, TransactionToken transaction,
             throw new TTransactionException();
         }
     }
+    
+    @Override
+    public Map<String, Map<Diff, Set<TObject>>> diffRecordStartEnd(long record,
+            long start, long end, AccessToken creds, TransactionToken transaction, String environment) throws TSecurityException, TTransactionException, TException {
+        PrettyLinkedTableMap<String, Diff, Set<TObject>> result = PrettyLinkedTableMap
```
jtnelson: need to call checkAccess(), etc at the beginning of each of these methods to validate the creds/transaction

concourse-server/src/main/java/org/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -1147,6 +1152,118 @@ public boolean commit(AccessToken creds, TransactionToken transaction,
...
+                .newPrettyLinkedTableMap();
+        result.setRowName("Value");
+        Map<String, Set<TObject>> startBrowse = selectRecordTime(record, start, creds, transaction, environment);
```
jtnelson: You shouldn't really call other ConcourseService methods in the server side implementations (unless its more efficient to do so) You should use the Engine/Transaction directly. In fact, you probably want to wrap these methods in an atomic operation. As good model for how to implement diff would be to look at the chronologize methods

thrift-api/concourse.thrift
```diff
@@ -948,6 +948,43 @@ service ConcourseService {
     7: string environment)
   throws (1: shared.TSecurityException ex, 2: shared.TTransactionException ex2);
```
jtnelson: you need a stub for `diffRecordStart`

thrift-api/concourse.thrift
```diff
@@ -948,6 +948,43 @@ service ConcourseService {
...
+    1: string key,
+    2: i64 record,
+    3: i64 start,
+    4: i64 tend,
+    5: shared.AccessToken creds,
+    6: shared.TransactionToken transaction,
+    7: string environment)
+  throws (1: shared.TSecurityException ex, 2: shared.TTransactionException ex2);
+
+  map<data.TObject, map<shared.Diff, set<i64>>> diffKeyStartEnd(
```
jtnelson: You'll also need `diffKeyStart`

jtnelson:
when you regenerated the thrift service, did the files for the ruby, php and python drivers change as well? Those should also get committed

PR #74, opened 2015-07-14
Added the Record.#recall method to fetch previous state of key/value …

dearyhud:
…pair N revisions ago.

@jtnelson my man - This is still an imperfect diff because, I'm getting back into this project. If you could help me understand how [Revisions](https://github.com/dearyhud/concourse/blob/014c05d28a0a6c7a11f5f879ebb4a99be851c17d/concourse-server/src/main/java/org/cinchapi/concourse/server/storage/db/Revision.java), and [Compact Revisions](https://github.com/dearyhud/concourse/blob/014c05d28a0a6c7a11f5f879ebb4a99be851c17d/concourse-server/src/main/java/org/cinchapi/concourse/server/storage/db/CompactRevision.java) work together, and how they fit in concourse's design (because I think those are new implementations) that would help me close this [issue] (https://cinchapi.atlassian.net/browse/CON-231) out a lot faster. You can hit me up on google hangouts or email. Thanks!

### DIFF
--- a/concourse-server/src/main/java/org/cinchapi/concourse/server/http/UIState.java
+++ b/concourse-server/src/main/java/org/cinchapi/concourse/server/http/UIState.java
@@ -28,8 +28,10 @@ import spark.Response;
 import com.google.common.collect.Maps;
 
 /**
- * A {@link UIState} is processed by a {@link Router} to display some front end
- * template.
+ * A {@link UIState} is an endpoint that is made up of a front end template
+ * (defined in the "templates" directory on the classpath) and some template
+ * variables. This should be used to return logic to render an initial UIState
+ * within the browser for the user.
  * 
  * @author Jeff Nelson
  * 


jtnelson:
Yo @dearyhud 

For this, you don't need to make any changes to the storage engine or the Record class directly. Here is how you should approach this change.

1. Add method stubs to thrift-api/concourse.thrift
recallKeyRecord
recallRecord

2. Run utils/compile-thrift-all.sh
This will generate the RPC code for the new methods in ConcourseServer and add method stubs in all the drivers.

3. Open ConcourseServer.java and implement the new methods
Eclipse will complain about you needing to implement the methods and will add the stubs for you. Then, you'll need to add the logic to implement the recall routine.

Once you get to step 3, lets chat about the specific implementation. Recall would need to be implemented as an AtomicOperation. There are a lot of examples that I can show you for how to do that

dearyhud:
yo @jtnelson - perfect. I went through said steps with thrift, and the stubs are in the drivers. Let me know when's a good time for you to chat.

PR #77, opened 2015-09-04
CON-259: Fix historical reads in the future

lhcopetti:
Historical reads by definition don't grab locks but if the timestamp was set to the future there was a possibility of a Phantom Read phenomenon.


jtnelson:
@lhcopetti can you sign the contributor license agreement at https://www.clahub.com/agreements/cinchapi/concourse

jtnelson:
This looks really good. After thinking about my suggestions to use the ternary operator more, I think you should use if/else statements like you were doing originally. Your code blocks would look something like:

```java
if(timestamp > Time.now()) {
    return select(record);
}
else {
    checkState();
    return super.select(record, timestamp);
}
```

The reason this is preferred is because it will ensure we don't call `checkState()` twice in the event that someone is using a future timestamp

lhcopetti:
You are definitely right. I missed that. I will proceed to do that as soon as I get home.  

Anything else just let me know...

PR #82, opened 2015-10-12
Invalid csv file

hcuffy:
These changes help to detect files that aren't csv in order to ensure importing the file is possible.


jtnelson:
It looks like your eclipse settings are changing the indentation levels. Can you do the following

1. Download the file at https://github.com/cinchapi/dev-tools/blob/master/CinchapiFormatter.xml
2. In Eclipse go to Preferences > Java > Code Style > Formatter and Import the file you downloaded 
3. For each file you changed, press SHIFT + CMD + F to reformat the code

concourse-integration-tests/src/test/resources/invalidLine.csv
```diff
@@ -0,0 +1 @@
+< I'm the invalid test file >
```
jtnelson: Can you rename this file to "invalid.csv"

concourse-integration-tests/src/test/java/org/cinchapi/concourse/importer/InvalidCsvFileFormatImportTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+	@Override
+	protected String getImportPath() {
+		return "invalidLine.csv";
+	}
+	
+	@Override
+	@Test (expected = IllegalArgumentException.class)
+	public void testImport() {
+        String file = Resources.get("/" + getImportPath()).getFile();
```
jtnelson: Blocks should be properly indented

concourse-import/src/main/java/org/cinchapi/concourse/importer/LineBasedImporter.java
```diff
@@ -47,240 +47,237 @@
...
-        }
-        return json;
-    }
+	/**
+	 * Validate if given file is csv
+	 * 
+	 * @param line this is the first line of the given file
+	 *            
+	 */
+	protected abstract void validateFileFormat(String line);
```
jtnelson: Since this is in an abstract class, the Javadoc must be generic since it can theoretically apply to more than just CSV files. Something like

"Check the {@code line} and determine if it is valid for the file format supported by this importer"

Also, change the comment about the line param being the "first" line to simple being "a" line in the file because its possible that we'll want to call this multiple times in the future to perform validation based on multiple lines.

Lastly, I would add a throws declaration to the method signature:
```java
protected abstract void validateFileFormat(String line) throws IllegalArgumentException;
```

And also add an the following line to the end of the javadoc:
@throws IllegalArgumentException if the line is not acceptable for the file format

hcuffy:
I updated the classes based on your feedback let me know if it's good.

PR #88, opened 2015-10-22
Cash should automatically convert method names in underscore case to camelcase enhancement

dubex:


concourse-integration-tests/src/test/java/org/cinchapi/concourse/shell/ConcourseShellTest.java
```diff
@@ -81,5 +81,12 @@ public void testSecurityChangeCausesCrash() throws Throwable {
         grantAccess("admin", "admin2");
         cash.evaluate("add \"name\", \"jeff\", 1");
     }
+    
+    @Test
+    public void testConvertMethodNamesInUnderscoreToCamelcase() throws IrregularEvaluationResult {
+    	String record = cash.evaluate("find_or_add('name', 'concourse')");
+    	 String result = cash.evaluate("get('name', " + record.split("'")[1] + ")");
```
jtnelson: Use consistent indentation

concourse-shell/src/main/java/org/cinchapi/concourse/shell/ConcourseShell.java
```diff
@@ -588,6 +591,25 @@ else if(e instanceof MissingMethodException
...
+                else if ((e.getMessage().contains(
+                        "No such property: ext for class: ConcourseShell") && !isSecond) || (e.getMessage().contains(
+                                "No signature of method: ConcourseShell.") && !isSecond)) {
+                	isSecond = true;
+                	if(e.getMessage().contains("ConcourseShell.")) {
+                		String method = e.getMessage().split("ConcourseShell.")[1]
+                                .split("\\(")[0];
+                		String methodWithCamecase = CaseFormat.LOWER_UNDERSCORE.to(CaseFormat.LOWER_CAMEL, method);
+                		input = input.replaceAll(method, methodWithCamecase);
+                	} else if(input.contains("ext")) {
```
jtnelson: else if should go on separate line

concourse-shell/src/main/java/org/cinchapi/concourse/shell/ConcourseShell.java
```diff
@@ -588,6 +591,25 @@ else if(e instanceof MissingMethodException
...
+                        String method = input.split("ext.")[1]
+                                .split("\\(")[0];
+                        input = input.replaceAll("ext." + method, method);
+                		input = CaseFormat.LOWER_UNDERSCORE.to(CaseFormat.LOWER_CAMEL, input);
+                	} else {
```
jtnelson: else should go on separate line

jtnelson:
Can you add another test to show that invoking a method that is not defined in an external script and is not underscore case of a valid method still throws an exception?

dubex:
Done as mentioned in commnet

dubex:
Fixed formatting in ConcourseShellTest

jtnelson:
Hey, my bad here. Need to make a slight change before I can merge. 

You added your logic to a separate else if block, but it should go into the existing block that handles MissingMethodException.

dubex:
Re factored the logic which handles the MissingMethodException

jtnelson:
Formatting is off

dubex:
Done.

jtnelson:
Can you also incorporate the fix for issue #91 into this PR?

dubex:
Done

jtnelson:
My mistake, I guided you wrong here :(

Turns out this approach won't work if there are multiple method calls in a single line. For example, if someone were to do the following (where "foo" and "bar" are methods defined in an external script):
```groovy
foo(); bar(); find_or_add("name", "concourse");
```
the logic would fail because it implicitly assumes that once, one method is converted, there is nothing left to do.

We shouldn't try to do the snake -> camel case conversion in the evaluate method. Instead, we should handle it in the same way that we do short syntax. Take a look at SyntaxTools.java. In particular the handleShortSyntax method. Study how that is called from ConcourseShell as well. Then lets chat about how you can take a similar approach to handleSnakeCase syntax for method invocations.

PR #109, opened 2015-11-22
Exit methods were modified in the connectionPool class and new unit t…

hcuffy:
…est methods were created to the connectionPoolTest class.

### DIFF
--- a/concourse-driver-java/src/main/java/com/cinchapi/concourse/ConnectionPool.java
+++ b/concourse-driver-java/src/main/java/com/cinchapi/concourse/ConnectionPool.java
@@ -52,6 +52,15 @@ import com.google.common.collect.Sets;
  * 
  * @author Jeff Nelson
  */
+
+/**
+ * @author cuffyhenry
+ *
+ */
+/**
+ * @author cuffyhenry
+ *
+ */
 @ThreadSafe
 public abstract class ConnectionPool implements AutoCloseable {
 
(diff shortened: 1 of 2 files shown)


concourse-driver-java/src/main/java/com/cinchapi/concourse/ConnectionPool.java
```diff
@@ -52,6 +52,15 @@
  * 
  * @author Jeff Nelson
  */
+
```
jtnelson: Delete these additional comment blocks

concourse-driver-java/src/main/java/com/cinchapi/concourse/ConnectionPool.java
```diff
@@ -446,11 +467,6 @@ private void exitAllConnections() {
                     exited = true;
                 }
                 catch (Exception e) {
-                    // If a shutdown hook is used to close the connection pool,
```
jtnelson: These comments need to be added back

concourse-driver-java/src/main/java/com/cinchapi/concourse/ConnectionPool.java
hcuffy: I erased and inserted the comments as per and committed the file.

Date: Sun, 22 Nov 2015 05:02:25 -0800

PR #126, opened 2016-01-01
Cash should automatically convert method names in underscore case to …

dubex:
…camelcase enhancement


concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
```diff
@@ -627,12 +651,24 @@ else if(e.getCause() instanceof SecurityException) {
...
                                     + "session cannot continue");
                 }
-                else if(e instanceof MissingMethodException
-                        && hasExternalScript()
-                        && ErrorCause.determine(e.getMessage()) == ErrorCause.MISSING_CASH_METHOD) {
-                    String method = e.getMessage().split("ConcourseShell.")[1]
-                            .split("\\(")[0];
-                    input = input.replaceAll(method, "ext." + method);
+                else if(e instanceof MissingMethodException) {
+                    String method = ((MissingMethodException) e).getMethod();
```
jtnelson: Does `MissingMethodException#getMethod()` return just the methodName or does it return ConcourseShell.methodName?

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
```diff
@@ -295,6 +296,21 @@ private static void die(String message) {
     }
 
     /**
+     * Convert all concourse API (or {@code methods}) to underscore case
+     * 
+     * @return
+     */
+    protected static Set<String> convertAccessibleApiMethodsToUnderscoreCase(
```
jtnelson: Use Underscore or UnderScore consistently in naming

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
```diff
@@ -360,6 +376,14 @@ private static String getHelpText(String topic) {
             .newArrayList(getAccessibleApiMethods());
 
     /**
+     * A list which contains all of the accessible API methods in underscore
+     * case. This list is
+     * used to detect the method in underscore case is supported by concourse.
+     */
+    private static final Set<String> methodsInUnderScoreCase = Sets
```
jtnelson: Use Underscore or UnderScore consistently in naming

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
```diff
@@ -295,6 +296,21 @@ private static void die(String message) {
     }
 
     /**
+     * Convert all concourse API (or {@code methods}) to underscore case
+     * 
+     * @return
```
jtnelson: Explain what the method returns

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
```diff
@@ -295,6 +296,21 @@ private static void die(String message) {
     }
 
     /**
+     * Convert all concourse API (or {@code methods}) to underscore case
```
jtnelson: Concourse should be capitalized

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
```diff
@@ -627,12 +651,24 @@ else if(e.getCause() instanceof SecurityException) {
...
-                        && hasExternalScript()
-                        && ErrorCause.determine(e.getMessage()) == ErrorCause.MISSING_CASH_METHOD) {
-                    String method = e.getMessage().split("ConcourseShell.")[1]
-                            .split("\\(")[0];
-                    input = input.replaceAll(method, "ext." + method);
+                else if(e instanceof MissingMethodException) {
+                    String method = ((MissingMethodException) e).getMethod();
+
+                    if(ErrorCause.determine(e.getMessage()) == ErrorCause.MISSING_CASH_METHOD
+                            && methodsInUnderScoreCase.contains("concourse."
```
jtnelson: It's possible that someone might call the method with "concourse." already prepended
For example, they could do
```groovy
concourse.find_or_add
```
In which case, here you'd be checking for concourse.concourse.find_or_add.

I'd recommend writing a helper function in `com.cinchapi.concourse.util.Strings` called `prependIfAbsent` that will only prepend a string (a) to the beginning of another string (b) if b does not already begin with a

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
```diff
@@ -627,12 +651,24 @@ else if(e.getCause() instanceof SecurityException) {
                             "A security change has occurred and your "
                                     + "session cannot continue");
                 }
-                else if(e instanceof MissingMethodException
```
jtnelson: should check if 
-e instanceof MissingMethodException
- error cause is missing cash method
- hasExternalScript OR method is valid underscore case

otherwise just let the exception fall through

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
```diff
@@ -627,12 +651,24 @@ else if(e.getCause() instanceof SecurityException) {
...
+                                    + method)) {
+                        String methodWithCamecase = CaseFormat.LOWER_UNDERSCORE
+                                .to(CaseFormat.LOWER_CAMEL, method);
+                        input = input.replaceAll(method, methodWithCamecase);
+                    }
+                    else if(hasExternalScript()
+                            && ErrorCause.determine(e.getMessage()) == ErrorCause.MISSING_CASH_METHOD) {
+                        input = input.replaceAll(method, "ext." + method);
+                    }
+                    else {
```
jtnelson: This block shouldn't be here. See the comment I made above about letting the exception fall through

jtnelson:
What if, instead of converting all the method names to underscore at the beginning, you do something like this:

- when a MissingMethodException is thrown, you convert the name of that method to from underscore to camel case and if the camel case version is in `methods` then you replace it in the line.

You could also define some static helper function to contain this logic
``java
private static isValidUnderscoreMethod(String method){
    // convert the method from underscore to camel case and then check the collection of methods to see if it is contained
}
```

**Comment by dubex** [comment]
Sure, it's much better option. I don't know I missed it.

File: concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java [hunk 9]
```diff
@@ -295,6 +296,15 @@ private static void die(String message) {
     }
 
     /**
+     * Checks API name in underscore is supported by Concourse. Returns true if supported else false.
+     * 
+     * @return
```
jtnelson: explain what is returned

dubex:
done.

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
dubex: It would return only methodName

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
dubex: If i remove else bolck the following test case will run into infinite loop.
```
public void testConvertWrongMethodNamesInUnderscoreToCamelcase()
            throws IrregularEvaluationResult {
        cash.evaluate("find_or_sub('name', 'concourse')");
 }
```

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
jtnelson: I would write it like this:
```java
else if(e instanceof MissingMethodException && ErrorCause.determine(e.getMessage()) == ErrorCause.MISSING_CASH_METHOD && (hasExternalScript() || isValidUnderscoreMethod(method)) {
    if(hasExternalScript()){
        // handle case when calling method from external script
    }
    else {
        // handle case when converting underscore to camel case
    }
}
```

jtnelson: The reason this way is preferred is because, you don't have to add the else block where you manually throw the EvaluationException. If the conditions above are not matched, then we shouldn't try to do any analysis on the exception and we should just let it fall through to the else statement that is already there on line 661 which means we aren't duplicating any code.

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
jtnelson: Cool! Didn't realize that!

concourse-shell/src/main/java/com/cinchapi/concourse/shell/ConcourseShell.java
```diff
@@ -627,12 +638,24 @@ else if(e.getCause() instanceof SecurityException) {
...
-                        && hasExternalScript()
-                        && ErrorCause.determine(e.getMessage()) == ErrorCause.MISSING_CASH_METHOD) {
-                    String method = e.getMessage().split("ConcourseShell.")[1]
-                            .split("\\(")[0];
-                    input = input.replaceAll(method, "ext." + method);
+                else if(e instanceof MissingMethodException) {
+                    String method = ((MissingMethodException) e).getMethod();
+
+                    if(ErrorCause.determine(e.getMessage()) == ErrorCause.MISSING_CASH_METHOD
+                            && isValidUnderscoreMethod("concourse."
```
jtnelson: 1. I would do the prepending of concourse. within the isValidUnderscoreMethod itself instead of prepending it to the parameter that is passed
2. You still need to account for the case if the called method already has concourse. prepended. For example, a user can call 
```groovy
concourse.find_or_add
```
and that is valid. But your logic will pass `concourse.concourse.find_or_add` to the isValidUnderscoreMethod

My suggestion is to create a helper function in `com.cinchapi.concourse.utl.Strings` that prepends a string to another string if and only if the other string does not already start with the string you want to prepend. You can call it something like `prependIfAbsent`

dubex:
Modified the code as suggested. Thanks for your suggestion.

PR #148, opened 2016-05-17
CON-158

hyinnn:
I want to mention that I'm getting a compiler warning for the varargs method.
"Type safety: Potential heap pollution"
I've tried to suppress it with "@SafeVarargs" but the compiler then says "Annotation cannot be applied to non-final instance methods". I can't figure out how to handle this, do you have any suggestions?

My reconcile logic was to remove any existing value that wasn't in the input and add the remaining input.

Please let me know if there are anything that needs to be changed.
Cheers


concourse-driver-java/src/main/java/com/cinchapi/concourse/Concourse.java
```diff
@@ -2193,6 +2194,28 @@ public abstract String jsonify(long record, Timestamp timestamp,
...
+    		Iterable<T> values);
+    
+    /**
+     * The {@code value} in {@code key} of {@code record} are 
+     * added and removed to be exactly the same as the input values
+     * 
+     * @param key the field name
+     * @param record the record id
+     * @param values var args of values
+     */
```
jtnelson: I think you can add `@SuppressWarnings("unchecked")` to suppress the compiler warning

interface/concourse.thrift
```diff
@@ -1963,6 +1963,18 @@ service ConcourseService {
     1: exceptions.SecurityException ex,
     2: exceptions.TransactionException ex2,
     3: exceptions.InvalidArgumentException ex3);
+
```
jtnelson: can you add a docblock here

concourse-driver-java/src/main/java/com/cinchapi/concourse/Concourse.java
```diff
@@ -5011,6 +5034,33 @@ public Boolean call() throws Exception {
...
+        }
+
+        @Override
+        public <T> void reconcile(final String key, final long record,
+                final Iterable<T> values) {
+            execute(new Callable<Void>() {
+
+                @Override
+                public Void call() throws Exception {
+                    Set<TObject> valueSet = Sets.newHashSet();
```
jtnelson: you should first check to see if the values collection is already a set. Check out `Collections#toList` method (in concourse-driver-java). I'd recommend you mimic that and create a `Collections#toSet` method that does a similar thing by checking if the collection is a set or not already.

concourse-driver-java/src/main/java/com/cinchapi/concourse/Concourse.java
```diff
@@ -5011,6 +5034,33 @@ public Boolean call() throws Exception {
...
+                    for (T value : values) {
+                        if (!valueSet.add(Convert.javaToThrift(value))) {
+                            throw new IllegalArgumentException(
```
jtnelson: I'd argue you don't need to throw an exception. You can let the set drop the duplicate values.

concourse-driver-java/src/main/java/com/cinchapi/concourse/Concourse.java
```diff
@@ -5011,6 +5034,33 @@ public Boolean call() throws Exception {
         }
 
         @Override
+        public <T> void reconcile(final String key, final long record,
+                final T... values) {
+            reconcile(key, record, Lists.newArrayList(values));
```
jtnelson: you should pass in a Set instead of a List since ultimately the thrift method requires a Set

concourse-driver-java/src/main/java/com/cinchapi/concourse/Concourse.java
```diff
@@ -5011,6 +5034,33 @@ public Boolean call() throws Exception {
...
+        }
+
+        @Override
+        public <T> void reconcile(final String key, final long record,
+                final Iterable<T> values) {
```
jtnelson: make values a collection instead of an iterable

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -2830,6 +2830,36 @@ public boolean pingRecord(long record, AccessToken creds,
...
+            AccessToken creds, TransactionToken transaction,
+            String environment) throws TException {
+        checkAccess(creds, transaction);
+        AtomicSupport store = getStore(transaction, environment);
+        AtomicOperation atomic = null;
+        while (atomic == null || !atomic.commit()) {
+            atomic = store.startAtomicOperation();
+            try {
+                Set<TObject> existingValues = 
+                        getStore(transaction, environment).select(key, record);
```
jtnelson: you should use the `store` variable you created previously

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -2830,6 +2830,36 @@ public boolean pingRecord(long record, AccessToken creds,
...
+                for (TObject existingValue : existingValues) {
+                    if (!values.remove(existingValue)) {
+                        removeKeyValueRecord(key, existingValue, record,
```
jtnelson: you should use the `atomic` operation to remove the value
```java
atomic.remove(key, existingValue, record)
```

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -2830,6 +2830,36 @@ public boolean pingRecord(long record, AccessToken creds,
...
+                                creds, transaction, environment);
+                    }
+                }
+                for (TObject value : values) {
+                    addKeyValueRecord(key, value, record, creds,
```
jtnelson: Same thing as above, using the `atomic` operation to make modifications

interface/concourse.thrift
hyinnn: sorry I'm not exactly sure what a docblock is

concourse-driver-java/src/main/java/com/cinchapi/concourse/Concourse.java
hyinnn: I'm unable to make the toSet method call because I have to convert each value to a TObject

jtnelson: ahh good point...what you're doing is correct then

I'd just again take in a collection instead of iterable and do
`Set<TObject> valueSet = Sets.newHashSetWithExpectedSize(values.size());`

hyinnn: done!

jtnelson:
@hyinnn just a few more things to do in order to merge this one:

1. Change the thrift method name to reconcileKeyRecordValues. There is so magic that happens in the driver methods that rely on having the method arguments being part of the thrift name
2. Add an implementation for reconcoleKeyRecordValues in Mockcourse.groovy so the unit tests for the php, ruby and python drivers will pass

Hit me up if you need help with either of those things

PR #169, opened 2016-06-10
Feature/con 451

RaghavBabu:
enable/disable feature added. and  error fixed in ReflectionTest.

### DIFF
--- a/concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
+++ b/concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
@@ -20,10 +20,7 @@ import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.channels.FileChannel;
 import java.security.SecureRandom;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -206,6 +203,12 @@ public class AccessManager {
     private final AccessTokenManager tokenManager;
 
     /**
+     * Set which holds disabled users
+     */
+    private Set<ByteBuffer> disabledUsers = new HashSet<ByteBuffer>();
+
+
+    /**
      * Construct a new instance.
      * 
      * @param backingStore
@@ -286,6 +289,37 @@ public class AccessManager {
         }
         finally {
             lock.unlockWrite(stamp);
+        }
+    }
+
+    /**
+     * Iterates through all the disabledUsers and if username matches the input username,
+     * removes it from the disabledUsers set.
+     *
+     * @param username
+     */
+    public void enableUser(ByteBuffer username) {
+        if(disabledUsers.contains(username)) {
+            Iterator<ByteBuffer> it = disabledUsers.iterator();
+            while (it.hasNext()) {
+                ByteBuffer user = it.next();
+                if (username.equals(user)) {
+                    it.remove();
+                    break;
+                }
+            }
+        }
+    }
+
+    /**
+     * Check if the user exists and if it matches,
+     * adds its to a disabledUsers set.
+     *
+     * @param username
+     */
+    public void disableUser(ByteBuffer username) {
+        if(credentials.containsValue(ByteBuffers.encodeAsHex(username))) {
+            disabledUsers.add(username);
         }
     }
 
@@ -548,6 +582,8 @@ public class AccessManager {
      *         .
      */
     private boolean isExistingUsername0(ByteBuffer username) {
+        if(disabledUsers.contains(username))
+            return false;
         return credentials.containsValue(ByteBuffers.encodeAsHex(username));
     }
 
--- a/concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
+++ b/concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
@@ -20,17 +20,16 @@ import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.channels.FileChannel;
 import java.security.SecureRandom;
-import java.util.*;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
-
 import jsr166e.StampedLock;
-
 import org.joda.time.format.DateTimeFormatter;
 import org.joda.time.format.DateTimeFormatterBuilder;
-
 import static com.google.common.base.Preconditions.*;
-
 import com.cinchapi.concourse.Timestamp;
 import com.cinchapi.concourse.annotate.Restricted;
 import com.cinchapi.concourse.server.io.FileSystem;
@@ -46,7 +45,6 @@ import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.Lists;
-import com.google.common.collect.Sets;
 import com.google.common.hash.Hashing;
 import com.google.common.primitives.Longs;
 
@@ -177,6 +175,12 @@ public class AccessManager {
      * table.
      */
     private static final String USERNAME_KEY = "username";
+    
+    /**
+     * The column that contains a boolean which indicates if a user is enabled or not {@link #credentials}
+     * table(When a user is created, its enabled by default).
+     */
+    private static final String USER_ENABLED = "user_enabled";
 
     /**
      * The store where the credentials are serialized on disk.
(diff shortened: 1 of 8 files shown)


concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -548,6 +582,8 @@ private void insert0(ByteBuffer username, ByteBuffer password,
      *         .
      */
     private boolean isExistingUsername0(ByteBuffer username) {
+        if(disabledUsers.contains(username))
```
jtnelson: add {

concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -548,6 +582,8 @@ private void insert0(ByteBuffer username, ByteBuffer password,
...
+            return false;
```
jtnelson: and closing brace `}` on next line

concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -206,6 +203,12 @@ protected static boolean isSecurePassword(ByteBuffer password) {
     private final AccessTokenManager tokenManager;
 
     /**
+     * Set which holds disabled users
+     */
+    private Set<ByteBuffer> disabledUsers = new HashSet<ByteBuffer>();
```
jtnelson: use `Sets.newHashSet` instead of creating a new HashSet directly.

jtnelson: I'd recommend maintaining a Set of strings instead of bytebuffers. Things can get messy with ByteBuffer positioning, etc.

concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -290,6 +293,37 @@ public void deleteUser(ByteBuffer username) {
     }
 
     /**
+     * Iterates through all the disabledUsers and if username matches the input username,
+     * removes it from the disabledUsers set.
+     *
+     * @param username
+     */
+    public void enableUser(ByteBuffer username) {
+        if(disabledUsers.contains(username)) {
```
jtnelson: You don't have to iterate though the set, you can just use the `contains` method

concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -290,6 +293,37 @@ public void deleteUser(ByteBuffer username) {
...
+    }
+
+    /**
+     * Check if the user exists and if it matches,
+     * adds its to a disabledUsers set.
+     *
+     * @param username
+     */
+    public void disableUser(ByteBuffer username) {
+        if(credentials.containsValue(ByteBuffers.encodeAsHex(username))) {
```
jtnelson: you should call `isExistingUsername0` here

concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -548,6 +582,8 @@ private void insert0(ByteBuffer username, ByteBuffer password,
      *         .
      */
     private boolean isExistingUsername0(ByteBuffer username) {
```
jtnelson: I thought about this some more. I think you should remove the change from isExistingUsername and create a new method `isEnabledUsername` that checks if its disabled. Any calls to isExistingUsername should also call isEnabledUsername as well.

jtnelson:
Use `StampedLock` to deal with concurrency like we discussed on Slack

concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -20,10 +20,7 @@
 import java.nio.CharBuffer;
 import java.nio.channels.FileChannel;
 import java.security.SecureRandom;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
```
jtnelson: Don't use static imports.

concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -290,6 +294,42 @@ public void deleteUser(ByteBuffer username) {
...
     /**
+     * Iterates through all the disabledUsers and if username matches the input username,
+     * removes it from the disabledUsers set.
+     *
+     * @param username
+     */
+    public void enableUser(ByteBuffer username) {
+        long stamp = lock.writeLock();
+        try {
+            if(disabledUsers.contains(username)) {
```
jtnelson: There is no need to do a contains check, especially since this method isn't returning a boolean. You can just call `disabledUsers.remove(username)`

concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -290,6 +294,42 @@ public void deleteUser(ByteBuffer username) {
...
+                disabledUsers.remove(username);
+            }
+        }
+        finally {
+            lock.unlockWrite(stamp);
+        }
+    }
+
+    /**
+     * Check if the user exists and if it matches,
```
jtnelson: This documentation could all probably fit on one line

concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -324,7 +364,7 @@ public void expireAccessToken(AccessToken token) {
      */
     public AccessToken getNewAccessToken(ByteBuffer username) {
         long stamp = lock.tryOptimisticRead();
-        checkArgument(isExistingUsername0(username));
+        checkArgument(isExistingUsername0(username) && isEnabledUsername(username));
```
jtnelson: this should call isEnabledUsername0

jtnelson: If your isEnabledUsername method is going to also check if the username exists, then you should just call isEnabledUsername0 here instead of calling both methods

concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -418,6 +458,29 @@ public boolean isExistingUsername(ByteBuffer username) {
...
+            stamp = lock.readLock();
+            try {
+                valid = isExistingUsername0(username);
+                enabled = isEnabledUsername0(username);
+            }
+            finally {
+                lock.unlockRead(stamp);
+            }
+        }
+        return valid && enabled;
```
jtnelson: this return statement should be contained within the try block

concourse-server/src/main/java/com/cinchapi/concourse/server/jmx/ConcourseServerMXBean.java
```diff
@@ -147,4 +147,16 @@
      */
     public void revoke(byte[] username);
 
+    /**
+     * Enable the user.
```
jtnelson: For the documentation, explain what enabling the user means (i.e. the user can be authenticated with the correct password).

concourse-server/src/main/java/com/cinchapi/concourse/server/jmx/ConcourseServerMXBean.java
```diff
@@ -147,4 +147,16 @@
...
+     * @param username
+     */
+    public void enableUser(byte[] username);
+
+    /**
+     * Disable the user.
```
jtnelson: For the documentation, explain what disabling the user means (i.e. the user cannot be authenticated for any purposes, even with the correct password)

concourse-server/src/test/java/com/cinchapi/concourse/security/AccessManagerTest.java
```diff
@@ -268,6 +268,34 @@ public void testRevokeUser() {
...
+    public void testNewlyCreatedUserIsEnabled() {
+        ByteBuffer username = getAcceptableUsername();
+        ByteBuffer password = getSecurePassword();
+        manager.createUser(username, password);
+        Assert.assertTrue(manager.isEnabledUsername(username));
+    }
+
+
+    @Test
     public void testIsValidUsernameAndPassword() {
```
jtnelson: Add a test to ensure that any previously granted access tokens for a disabled user are no longer valid

concourse-server/src/main/java/com/cinchapi/concourse/security/AccessManager.java
```diff
@@ -290,6 +294,42 @@ public void deleteUser(ByteBuffer username) {
...
+     * adds its to a disabledUsers set.
+     *
+     * @param username
+     */
+    public void disableUser(ByteBuffer username) {
+        long stamp = lock.writeLock();
+        try {
+            if (isExistingUsername0(username)) {
+                disabledUsers.add(username);
+            }
```
jtnelson: You'll need to also delete all the access tokens the user has. See how the deleteUser method makes a call to tokenManager.deleteAllUserTokens

PR #174, opened 2016-06-14
CON-468, inventory() method will return results as expected.

RaghavBabu:


concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -2774,7 +2774,7 @@ public boolean insertJsonRecord(String json, long record,
     public Set<Long> inventory(AccessToken creds, TransactionToken transaction,
             String environment) throws TException {
         checkAccess(creds, transaction);
-        return getEngine(environment).browse();
+        return getEngine(environment).getAllRecords();
```
jtnelson: this will return the same thing, you need to do
getStore(environment, transaction) and call getAllRecords on that

jtnelson:
Please add unit test. Whenever fixing a bug, add a unit test to first repro the bug and then verify its fixed. You can add a unit test in the concourse-integration-test project in the com.cinchapi.concourse.bugrepro package. 

The actual repro can be the steps I outlined in the jira ticket

Issue #115, opened 2015-11-29
Order of params in link method is unintuitive


jtnelson:
The `add`, `set` and `remove` methods have a signature with order `key, value, record`
The `link` and `unlink` methods have a signature with order `key, source, destination`

it would probably be more intuitive to have the link signature be `key, destination, source`

PR #175, opened 2016-06-15
GH-115: Changed the parameter order for 'link' and 'unlink' method

javierlores:
The parameters for the two 'link' methods and 'unlink' method in the class Concourse were changed from (key, source, destination) to (key, destination, source). This was done to match the parameter order of (key, value, record) for the add, set, and remove methods and thus make the parameter order more intuitive. All usages were modified to adjust for this change.


javierlores:
@RaghavBabu Check my commit if you could please

PR #180, opened 2016-06-21
CON-321: Added support for local references in CCL

javierlores:
In the class Parser, local reference support into a Multimap from a CCL string was added within the toPostfixNotation() method. This is needed to allow users to insert data and reference a blob of JSON data. For example insert(blob, "ipeds_id = $ipeds_id") would replace $ipeds_id by values within the blob that match ipeds_id
Unit tests were added to verify the functionality of this new update toPostfixNotation method.

### DIFF
--- a/concourse-server/src/main/java/com/cinchapi/concourse/lang/Parser.java
+++ b/concourse-server/src/main/java/com/cinchapi/concourse/lang/Parser.java
@@ -211,10 +211,6 @@ public final class Parser {
                 SplitOption.TOKENIZE_PARENTHESIS);
         List<Symbol> symbols = Lists.newArrayList();
         GuessState guess = GuessState.KEY;
-        boolean shouldResolveVariableValue = false;
-        KeySymbol resolvedVariableKey = null;
-        OperatorSymbol resolvedVariableOperator = null;
-        ConjunctionSymbol resolvedVariableConjunction = null;
         StringBuilder buffer = null;
         StringBuilder timeBuffer = null;
         while (toks.hasNext()) {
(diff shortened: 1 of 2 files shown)


concourse-server/src/main/java/com/cinchapi/concourse/lang/Parser.java
```diff
@@ -226,23 +249,130 @@ else if(StringUtils.isBlank(tok)) {
...
+                    if(tok.length() > 2 && tok.substring(0, 2).equals(("$$"))) {
+                        tok = tok.substring(2);
+                        conjunctionRes = ConjunctionSymbol.AND;
+                    }
+                    else if(tok.length() > 1 && tok.substring(0, 1).equals("$")) {
+                        tok = tok.substring(1);
+                        conjunctionRes = ConjunctionSymbol.OR;
+                    }
+                    varRes = true;
+                    symbols.add(ParenthesisSymbol.parse("("));
```
jtnelson: you can add `ParenthesisSymbol.LEFT` directly

concourse-server/src/main/java/com/cinchapi/concourse/lang/Parser.java
```diff
@@ -226,23 +249,130 @@ else if(StringUtils.isBlank(tok)) {
...
+                    keyRes = KeySymbol.parse(tok);
+                    symbols.add(keyRes);
+                    guess = GuessState.OPERATOR;
```
jtnelson: You should pull setting `guess` outside the if-else block since its the same in both cases

concourse-server/src/main/java/com/cinchapi/concourse/lang/Parser.java
```diff
@@ -192,6 +211,10 @@ else if(symbol == ParenthesisSymbol.RIGHT) {
                 SplitOption.TOKENIZE_PARENTHESIS);
         List<Symbol> symbols = Lists.newArrayList();
         GuessState guess = GuessState.KEY;
+        boolean varRes = false;
```
jtnelson: Since these variables hold different kind of values, I'd use different name schemes for clarity

varRes => shouldResolveVariableValue
keyRes => resolvedVaribleKeyVariable
operatorsRes => resolvedVariableOperator
conjunctionRes => resolvedVariableConjunction

concourse-server/src/main/java/com/cinchapi/concourse/lang/Parser.java
```diff
@@ -226,7 +249,23 @@ else if(StringUtils.isBlank(tok)) {
                 continue;
             }
             else if(guess == GuessState.KEY) {
-                symbols.add(KeySymbol.parse(tok));
+                if(tok.charAt(0) == '$') {
```
jtnelson: You shouldn't use `String#substring` unless you actually plan to use it because it creates a new String. You really want to do something like this:
```java
if(tok.charAt(0) == '$') {
    if(tok.length > 2 && tok.charAt(1) == '$') {
        // do logic for $$key
    }
    else {
        // do logic for $key
    }
}

jtnelson:
manually merged

Issue #182, opened 2016-06-23
CaSH erroneously converts key names with underscores to camel case


jtnelson:
```java
[default/cash]$ select 1461262580169000
Returned '
+---------------------------------------------------------------------------------------------+
| Key          | Values                                                                       |
+---------------------------------------------------------------------------------------------+
| fav_language | [Go]                                                                         |
| first_name   | [Jeff]                                                                       |
| friends      | [@1461262580175003, @1461262580175007, @1461262580178001, @1461262580179003] |
| last_name    | [Nelson]                                                                     |
+---------------------------------------------------------------------------------------------+
' in 0.048 sec

[default/cash]$ browse "fav_language"
Returned '
+-----------------------+
| favLanguage | Records |
+-----------------------+
+-----------------------+
' in 0.015 sec
```

saagar2000:
Can I take a look at this one?

PR #184, opened 2016-06-24
GH-182.. Modified logic such that only underscore in command is

RaghavBabu:
GH-182.. Modified logic such that only underscore in command is  …
converted to camel case.

### DIFF
--- a/concourse-shell/src/main/java/com/cinchapi/concourse/shell/SyntaxTools.java
+++ b/concourse-shell/src/main/java/com/cinchapi/concourse/shell/SyntaxTools.java
@@ -139,13 +139,14 @@ public final class SyntaxTools {
                 }
                 return line;
             }
-            else {	
+            else {
                 String[] query = line.split("\\s+");
                 String cmd = query[0];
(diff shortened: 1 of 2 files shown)


concourse-shell/src/main/java/com/cinchapi/concourse/shell/SyntaxTools.java
```diff
@@ -140,8 +140,11 @@ else if(!line.contains("(")) {
                 return line;
             }
             else {	
-            	if(line.contains("_")) { //CON-457
-                    line = CaseFormat.LOWER_UNDERSCORE.to(CaseFormat.LOWER_CAMEL, line);
+                String[] query = line.split("\\s+");
+                String cmd = query[0];
```
jtnelson: line 144 and 145 don't look aligned

jtnelson:
@RaghavBabu your unit test is likely okay, but it is good practice to add a repro test that follows the exact workflow mentioned in the bug so we don't run the risk of missing any other subtleties.

PR #185, opened 2016-06-25
[CON-325]: Added counters for verifications initiated and scanned, and calculated percentage

as577:
In order to calculate the percentage of verifications that actually scanned the buffer, I created three variables: 

(1) the number of verifications initiated
(2) number of verifications that scanned the buffer
(3) the percentage of verifications that scanned the buffer

To calculate (1), I added code to the `verify()` method within `Buffer.java` to increment the counter every time the for-loop was iterated through.

To calculate (2), I added code to the `pageMightContainRelevantWrites()` method within `WriteSeekingIterator` to increment the counter every time the method returned `true`.

Finally, (3) was set to the quotient of (1) and (2) in the constructor of the `Buffer`.


concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -358,6 +373,7 @@ public Buffer(String directory) {
                                                  // there is no call to
                                                  // #setInventory
         this.threadNamePrefix = "buffer-" + System.identityHashCode(this);
+        this.percentageOfVerifiesScanningBuffer = ((double) this.numberOfVerifiesTouchingPage)/this.numberOfVerifiesInitiated;
```
jtnelson: this value is only set once, its not going to be updated when the numerator and denominator change.

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -1176,25 +1193,28 @@ public void remove() {
          */
         public boolean mightContain(Write write) {
             Type valueType = write.getValue().getType();
-            if(writeCache.mightContainCached(write.getRecord(), write.getKey(),
```
jtnelson: You are using an inconsistent code style here. In one pace you do
```java
if(mightContain) {
    numberOfVerifiesTouchingPage++;
}
```
and others you do
```java
int increment = mightContain ? 1 : 0;
 numberOfVerifiesTouchingPage += increment;
```
You should pick one style or the other

jtnelson: also, it seems it would be easier to simply add logic to the WriteSeekingIterator#pageMightContainRelevantWrites method so you only need to make the change in one place

jtnelson: You'll also need to ensure you only count once when a page scans the buffer. During a single verify, pageMightContainRelevantWrites is likely to return true more than once. But you should only increment the variable for tracking actual scans by 1. You'll need to control for this by checking to see if the variable is already incremented from 0 before you do it again.

I'd suggest adding some variable to the WriteSeekingIterator that tracks whether it actually perform a scan. If the variable is already set to true then the value for buffer scans should not be incremented any more

as577:
@jtnelson I made some revisions based on your feedback.

(1) I changed the variables to JavaFX Properties and Bindings Objects, which allowed me to bind the percentage to the counters easily so as to have the quotient update automatically.

(2) I moved the logic from `verify()` to `mightContain()`, and made the code style more consistent.

(3) I added a `boolean scanned` which is set to whatever `mightContain()` returns. So, if `mightContain()` returns `true`, then the buffer was actually scanned and `scanned = true`. Incrementing counters only occurs if `scanned` is not true, so this prevents multiple increments to the variables if a scan has already been detected. **I had a question about this: should the number of verifies initiated also only be incremented once per a single verify? The way it is written now assumes this.**

as577:
Also, I attempted to devise a test for this but was unable to do so. I printed the value of the percentage for other test cases in `BufferTest` and saw values within the acceptable range (0 - 1), but I'm not sure how to verify whether these values are correct.

jtnelson:
It would be easier to have have a method to get the percent of verifies that scan as opposed to introducing a dependency on JavaFX.

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -1176,25 +1206,48 @@ public void remove() {
...
-                    write.getValue())) {
-                return true;
-            }
-            else if(valueType == Type.STRING) {
-                return writeCache.mightContainCached(write.getRecord(), write
-                        .getKey(), Value.wrap(Convert.javaToThrift(Tag
-                        .create((String) write.getValue().getObject()))));
+            incrementCounter(numberOfVerifiesInitiated, 1);
```
jtnelson: I'm not sure whats going on here, but the formatting looks off so its hard for me to see whats actually change. In any case, i don't think you need to make any changes here.

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -1176,25 +1193,28 @@ public void remove() {
          */
         public boolean mightContain(Write write) {
             Type valueType = write.getValue().getType();
-            if(writeCache.mightContainCached(write.getRecord(), write.getKey(),
-                    write.getValue())) {
-                return true;
+            boolean mightContain = writeCache.mightContainCached(write.getRecord(), write.getKey(),
+                    write.getValue());
+            if(mightContain) {
```
jtnelson: You still have this inconsistency (see below)

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -1176,25 +1193,28 @@ public void remove() {
...
+            	numberOfVerifiesTouchingPage++;
             }
             else if(valueType == Type.STRING) {
-                return writeCache.mightContainCached(write.getRecord(), write
+            	mightContain = writeCache.mightContainCached(write.getRecord(), write
                         .getKey(), Value.wrap(Convert.javaToThrift(Tag
                         .create((String) write.getValue().getObject()))));
+            	int increment = mightContain ? 1 : 0;
+            	numberOfVerifiesTouchingPage += increment;
```
jtnelson: You still have this inconsistency (see above)

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -741,6 +757,7 @@ public void transport(PermanentStore destination, boolean sync) {
     @Override
     public boolean verify(Write write, long timestamp, boolean exists) {
         for (Iterator<Write> it = iterator(write, timestamp); it.hasNext();) {
+        	this.numberOfVerifiesInitiated++;
```
jtnelson: why is this being incremented inside the for loop? It should happen before the for loop so its only called once per verify

jtnelson:
@as577 gonna hit you up on slack about this

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -73,6 +73,12 @@
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 
+import javafx.beans.binding.NumberBinding;
```
jtnelson: remove these imports since they are not being used

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -114,7 +120,17 @@
      * The average number of bytes used to store an arbitrary Write.
      */
     private static final int AVG_WRITE_SIZE = 30; /* arbitrary */
-
+    
+    /**
+     * The number of verifies initiated.
+     */
+    private AtomicLong numberOfVerifiesInitiated;
```
jtnelson: rename this `numVerifies`

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -114,7 +120,17 @@
...
+    
+    /**
+     * The number of verifies scanning the buffer.
+     */
+    private AtomicLong numberOfVerifiesTouchingPage;
```
jtnelson: rename this `numVerifyScans`

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -740,6 +758,7 @@ public void transport(PermanentStore destination, boolean sync) {
 
     @Override
     public boolean verify(Write write, long timestamp, boolean exists) {
+    	this.numberOfVerifiesInitiated.incrementAndGet();
```
jtnelson: dont use `this.variable` when there is no ambiguity. Just use `variable`

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -1632,7 +1657,19 @@ protected WriteSeekingIterator(Write write, long timestamp) {
...
+        	boolean mightContain = page.mightContain(write);
+        	if(!scanned && mightContain) {
+        		numberOfVerifiesTouchingPage.incrementAndGet();
+        	}
+        	return mightContain;
+        }
+        
+        /**
+         * Returns the percentage within range [0, 1] of verifies that scan the buffer.
+         * @return
```
jtnelson: describe what is returned for `@return`

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -1632,7 +1651,20 @@ protected WriteSeekingIterator(Write write, long timestamp) {
 
         @Override
         protected boolean pageMightContainRelevantWrites(Page page) {
```
jtnelson: Your formatting looks off here.

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -1632,7 +1651,20 @@ protected WriteSeekingIterator(Write write, long timestamp) {
...
+        	return mightContain;
+        }
+        
+        /**
+         * Determines the percentage within range [0, 1] of verifies that scan the buffer.
+         * 
+         * @return: decimal percentage of verifies initiated that scanned the buffer.
+         */
+        private float getPercentVerifyScans() {
+        	return ((float) numVerifyScans.get())/numVerifyRequests.get();
```
jtnelson: formatting looks off here too

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/temp/Buffer.java
```diff
@@ -114,7 +114,17 @@
      * The average number of bytes used to store an arbitrary Write.
      */
     private static final int AVG_WRITE_SIZE = 30; /* arbitrary */
-
```
jtnelson: Apply formatter to this file

jtnelson:
@as577 this is looking good! For testing, you should add a cases to BufferTest.java. I can think of the following two cases off the top of my head:

1. Add a random number of random objects to the buffer and then iterate over each of the objects to verify and finally assert that the percentVerifyScans is equal to 1.
2. Add a random number of random objects to the buffer and then add a random number of those objects to another set of other random objects and iterate over that set to verify and finally assert that the percentVerifyScans is greater than or equal to the percent of elements in the set that were originally added to the buffer

concourse-server/src/test/java/com/cinchapi/concourse/server/storage/temp/BufferTest.java
```diff
@@ -266,5 +268,57 @@ public void testPageExpansion() {
             GlobalState.BUFFER_PAGE_SIZE = oldBufferPageSize;
         }
     }
+    
+    @Test
+    public void testCompleteVerificationScan() {
+        Buffer buffer = (Buffer) store;
+        buffer.start();
+        addRandomElementsToBufferAndList(buffer, TestData.getScaleCount());
```
jtnelson: You should save the list to a variable

concourse-server/src/test/java/com/cinchapi/concourse/server/storage/temp/BufferTest.java
```diff
@@ -266,5 +268,57 @@ public void testPageExpansion() {
...
+        Iterator<Write> iterator = buffer.iterator();
```
jtnelson: You should iterate over the Writes in the list and not in the buffer directly

concourse-server/src/test/java/com/cinchapi/concourse/server/storage/temp/BufferTest.java
```diff
@@ -266,5 +268,57 @@ public void testPageExpansion() {
...
+            buffer.verify(iterator.next(), true);
+        }
+        float percent = Reflection.call(buffer, "getPercentVerifyScans");
+        Assert.assertTrue(percent == 1.0f);
+    }
+    
+    @Test
+    public void testComparePercentVerificationBetweenTwoSets() {
+        Buffer buffer = (Buffer) store;
+        buffer.start();
```
jtnelson: You don't have to manually start the buffer. The test framework does it automatically

concourse-server/src/test/java/com/cinchapi/concourse/server/storage/temp/BufferTest.java
```diff
@@ -266,5 +268,57 @@ public void testPageExpansion() {
             GlobalState.BUFFER_PAGE_SIZE = oldBufferPageSize;
         }
     }
+    
+    @Test
+    public void testCompleteVerificationScan() {
+        Buffer buffer = (Buffer) store;
+        buffer.start();
```
jtnelson: You don't have to manually start the buffer. The test framework does it automatically

concourse-server/src/test/java/com/cinchapi/concourse/server/storage/temp/BufferTest.java
```diff
@@ -266,5 +268,57 @@ public void testPageExpansion() {
...
+        addRandomElementsToBufferAndList(buffer, TestData.getScaleCount());
+        Iterator<Write> iterator = buffer.iterator();
+        while(iterator.hasNext()) {
+            buffer.verify(iterator.next(), true);
+        }
+        float percent = Reflection.call(buffer, "getPercentVerifyScans");
+        Assert.assertTrue(percent == 1.0f);
```
jtnelson: Use `Assert.assertEquals`

concourse-server/src/test/java/com/cinchapi/concourse/server/storage/temp/BufferTest.java
```diff
@@ -266,5 +268,57 @@ public void testPageExpansion() {
...
+     * the {@link Buffer} and a {@code List<Write>}, and returns the list.
+     * 
+     * @param buff: the buffer into which objects are inserted
+     * @param size: the number of objects to insert
+     * @return: a {@code List} of {@link Write} objects that were also inserted into the buffer
+     */
+    private List<Write> addRandomElementsToBufferAndList(Buffer buff, int size) {
+        List<Write> stored = Lists.newArrayList();
+        for(int i = 0; i < size; ++i) {
+            Write write = Write.add(TestData.getSimpleString(), TestData.getTObject(), i);
```
jtnelson: instead of adding to record `i`, add record `TestData.getLong()`

concourse-server/src/test/java/com/cinchapi/concourse/server/storage/temp/BufferTest.java
```diff
@@ -266,5 +268,57 @@ public void testPageExpansion() {
...
+        List<Write> stored2 = addRandomElementsToBufferAndList(buffer, TestData.getScaleCount());
+        int initialSize = stored2.size();
+        Random rand = new Random();
+        int transferSize = rand.nextInt(stored1.size());
+        for(int i = 0; i < transferSize; i++) {
+            stored2.add(stored1.get(i));
+        }
+        Iterator<Write> iterator = stored2.iterator();
+        while(iterator.hasNext()) {
+            buffer.verifyFast(iterator.next());
```
jtnelson: Why are you calling `verifyFast` here and not `verify`? Its not incorrect, per se, but it seems inconsistent

PR #186, opened 2016-06-27
[CON-328] Allow specifying the concourse.prefs file when starting the server

as577:
I implemented this feature by editing the concourse bash file and `GlobalState`.

In `concourse`, I extended the previous parsing of the `start` command to accept an additional `--config` flag. The variable `PREFS` was then set to the path specified after the `--config` flag if it was valid, otherwise it was set to the default preferences file.

In `GlobalState`, I set a variable `String prefs` to be equal to the property retrieved from the `concourse` script. Earlier, there were two variables which represented a developer preference file and a default preference file. Now, there is just one since the script takes care of defaulting.


concourse-server/scripts/concourse
```diff
@@ -568,10 +573,30 @@ status() {
   fi
 }
 
+# Sets the configuration/preferences file to the specified path
+config() {
+  PREFS=$APP_HOME"$1"
```
jtnelson: You should assume that the value for --config is going to be an absolute path

concourse-server/scripts/concourse
```diff
@@ -568,10 +573,30 @@ status() {
...
+
 # MAIN: Ensure that we are in the correct directory and do the requested action.
 cd $APP_HOME
 case "$1" in
   'start' | '--start')
+  shift
+  for var in "$@"; do
+    case "$var" in
+      '--config')
+      shift
```
jtnelson: this assume that `--config` will be the first flag, which isn't necessarily true

concourse-server/scripts/concourse
```diff
@@ -204,10 +204,15 @@ get_cli_for_alias(){
...
 }
 
-# The location of the concourse.prefs file that is used to configure the
+# The location of the default concourse.prefs file that is used to configure the
 # application
 PREFS=$APP_HOME"/conf/concourse.prefs"
 
+# The preference location is passed as a JVM option
+JVMOPTS=$JVMOPTS"
+-Dcom.cinchapi.concourse.server.prefs=$PREFS
```
jtnelson: the concatenated string will be bound to the variable in its current state. when you updated $PREFS later, this string won't also be updated

jtnelson:
1. Check to see if there is an arg parsing library for bash so you don't have to do it manually

jtnelson:
this is the order of precedence for potential prefs files:
1. Prefs passed in with --config
2. $APP_HOME/conf/concourse.prefs.dev if it exists
3. $APP_HOME/conf/concourse.prefs

concourse-server/src/main/java/com/cinchapi/concourse/server/GlobalState.java
```diff
@@ -177,16 +177,10 @@
...
-                PREFS_FILE_PATH = FileSystem.expandPath(devPrefs);
-            }
-            else {
-                config = ConcourseServerPreferences.open(defaultPrefs);
-                PREFS_FILE_PATH = FileSystem.expandPath(defaultPrefs);
-            }
+            String prefs = System.getProperty("com.cinchapi.concourse.server.prefs");
+            config = ConcourseServerPreferences.open(prefs);
+            PREFS_FILE_PATH = FileSystem.expandPath(prefs);
+            System.out.println("hello");
```
jtnelson: remove this

concourse-server/log/console.log
```diff
@@ -1 +1,6 @@
-
```
jtnelson: remove this file

PR #188, opened 2016-06-28
CON-474: repro test and bug fix

jtnelson:
Added logic to only add an empty snapshot to the database's context
if that particular snapshot is the last that the database knows
about


jtnelson:
@RaghavBabu @javierlores @as577 FYI

PR #195, opened 2016-07-28
[VIZ-7] Determine VariableType in TrackingMultimap

as577:


concourse-plugin-core/src/main/java/com/cinchapi/concourse/plugin/data/TrackingMultimap.java
```diff
@@ -192,6 +192,33 @@ public double uniqueness() {
...
+     * The three possible return types are:
+     * <ol>
+     *      <li><strong>DICHOTOMOUS</strong>: if there are 1 or 2 unique values</li>
+     *      <li><strong>NOMINAL</strong>: if the number of unique values is greater than 2 and less than or equal to 12</li>
+     *      <li><strong>INTERVAL</strong>: if there are more than 12 unique values</li>
+     * </ol>
+     * 
+     * @return
+     */
+    public VariableType variableType() {
```
jtnelson: fix formatting here...apply the CinchapiFormatter so the if/else blocks are correct

concourse-plugin-core/src/test/java/com/cinchapi/concourse/plugin/data/TrackingMultimapVariableTypeTest.java
```diff
@@ -0,0 +1,61 @@
+package com.cinchapi.concourse.plugin.data;
```
jtnelson: Can you add the Apache License header here

PR #200, opened 2016-07-30
[VIZ-15] Fixed implementation of #insert in Dataset and Unit Tested

as577:


concourse-plugin-core/src/main/java/com/cinchapi/concourse/plugin/data/Dataset.java
```diff
@@ -193,7 +193,7 @@ public boolean insert(E entity, A attr, V value) {
             entities = Sets.newHashSet();
             index.put(value, entities);
         }
-        if(entities.add(entity)) {
+        if(index.get(value).add(entity)) {
```
jtnelson: why'd you change this line?

as577: I'm not exactly sure why, but this fixed the glitchy implementation. If you revert it back and run the Unit Test, it fails.

as577: For some reason it wasn't adding it to the `Set<E>` that was associated with the map.

jtnelson: oh, i know why...cool

PR #202, opened 2016-07-31
Added support for CON API having both ( and space

dubex:


concourse-shell/src/main/java/com/cinchapi/concourse/shell/SyntaxTools.java
```diff
@@ -190,13 +195,14 @@ else if(!line.contains("(")) {
...
-                                   // for the blacklist
+                + ")(?<!\")[\\w\\.]+[\\(\\s]"; // match any word followed by an
+                                               // paren or whitespace char except
+        // for the blacklist
         Pattern pattern = Pattern.compile(regex);
         Matcher matcher = pattern.matcher(line);
         while (matcher.find()) {
             if(!matcher.group().startsWith("concourse.")) {
-                methods.add(matcher.group().replace("(", ""));
+                methods.add(matcher.group().replaceFirst("[\\(\\s]", ""));
```
jtnelson: why `relaceFirst` instead of `replace`

concourse-shell/src/main/java/com/cinchapi/concourse/shell/SyntaxTools.java
```diff
@@ -190,13 +195,14 @@ else if(!line.contains("(")) {
         Set<String> methods = Sets.newHashSet();
         Set<String> blacklist = Sets.newHashSet("time", "date");
         String regex = "\\b(?!" + StringUtils.join(blacklist, "|")
-                + ")[\\w\\.]+\\("; // match any word followed by an paren except
-                                   // for the blacklist
+                + ")(?<!\")[\\w\\.]+[\\(\\s]"; // match any word followed by an
+                                               // paren or whitespace char except
+        // for the blacklist
```
jtnelson: indent this line to align with the comments above

dubex: ok will do

concourse-shell/src/main/java/com/cinchapi/concourse/shell/SyntaxTools.java
dubex: replace intakes only char as an input. I guess replaceAll would work. I would make the chenges.

PR #203, opened 2016-08-01
[VIZ-18] Updated tracking method for unique value counts

as577:


concourse-plugin-core/src/main/java/com/cinchapi/concourse/plugin/data/TrackingMultimap.java
```diff
@@ -133,6 +133,11 @@ else if(clazz == Boolean.class || clazz == boolean.class) {
      * </p>
      */
     private final SparseBitSet valueCache;
+    
+    /**
+     * A {@link Set} to track the distinct keys entered into the {@link TrackingMultimap}.
+     */
+    private Set<K> keys;
```
jtnelson: isn't this the same as `data.keySet()`

PR #206, opened 2016-08-14
Created DatasetEntry class for encapsulating entries for the Dataset

as577:
In order for parts of the key-derivation to be detached from the process of updating the `Dataset`, I created a container that is instantiated by these parts and then passed back to the main `ImpromptuApplication` which takes care of entering them into the `Dataset`. Unit Tests are included.

### DIFF
--- a/concourse-plugin-core/src/main/java/com/cinchapi/concourse/plugin/data/Dataset.java
+++ b/concourse-plugin-core/src/main/java/com/cinchapi/concourse/plugin/data/Dataset.java
@@ -23,7 +23,6 @@ import java.util.Set;
 
 import javax.annotation.concurrent.NotThreadSafe;
 
-import com.cinchapi.concourse.plugin.data.entry.DatasetEntry;
 import com.cinchapi.concourse.server.plugin.io.PluginSerializable;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.Maps;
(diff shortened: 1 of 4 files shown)


concourse-plugin-core/src/main/java/com/cinchapi/concourse/plugin/data/Dataset.java
```diff
@@ -223,6 +223,33 @@ public boolean insert(E entity, A attribute, V value) {
...
+
+    /**
+     * Inserts a {@link Set<DatasetEntry>} to the {@link Dataset}
+     * 
+     * @param entry
+     * @return {@code true} if the association can be added because it didn't
+     *         previously exist
+     */
+    public boolean insert(Set<DatasetEntry<E, A, V>> entries) {
+        for (DatasetEntry<E, A, V> entry : entries) {
```
jtnelson: Is this supposed to be atomic such that either all entries get added or none at all? For example, What should happen if I have 10 entires, and the 4th one already exists?

jtnelson:
Whats the advantage of Impromptu creating a DatasetEntry and adding that to the data set as opposed to adding E, A, V individually using the regular `insert` methods?

concourse-plugin-core/src/main/java/com/cinchapi/concourse/plugin/data/Dataset.java
as577: Just submitted a pull request testing this. It won't insert the duplicate but it will insert the rest.

as577:
The rationale for having a container is that for deriving keys, I have particular classes that derive the keys and match them to their appropriate values and records. However, I did not want these classes to have access to the dataset since they needed write-only access. Using the container ensures that these classes merely make these associations and pass them back to the class that originally had the dataset which can perform the insertion. The Dataset is an important data structure that should be accessed by as few classes as possible.

I could probably have used a nested Map as a container and used that, but that would have convoluted the logic and required more boilerplate code to instantiate

PR #207, opened 2016-08-16
Created Insertable interface

as577:


concourse-plugin-core/src/main/java/com/cinchapi/concourse/plugin/data/Insertable.java
```diff
@@ -0,0 +1,34 @@
+package com.cinchapi.concourse.plugin.data;
```
jtnelson: can you add the open source apache license header

PR #209, opened 2016-08-30
[CON-496]  Concourse / Task CON-496 Return Dataset from ConcourseServer select methods that return Map<Long, Map<String, Set<TObject>>>

as577:
I changed the `ResultDataset` class to be type parameterized. This is because, by default, `ResultDataset` extended `Dataset<Long, String, Object>`, which made it incompatible to cast to a `Map<Long, Map<String, Set<TObject>>>`. By type parameterizing, this can be specified. If I were to explicitly specify that `ResultDataset` extended `Dataset<Long, String, TObject>`, the problem would be solved for `ConcourseServer` but there would then be incompatibilities with unit testing for the visualization engine and impromptu which use Objects instead of TObjects. This solution therefore maintains backwards compatibility.

Since `RemoteInvocationThread` is a package-friendly class, I was unable to call `instanceof` and instead used `Reflection.getCastedClass(String className)` to get the `Class<?>`, which I then used to determine if the `Thread.currentThread()` was assignable from it. The benefit of this approach is that there is no need to increase the visibility of `RemoteInvocationThread`, but it does create a hidden dependency due to the explicit package name specification in the `Reflection` call. Another solution would be to link the `Reflection` call to another related class in the same package as `RemoteInvocationThread`, such that in the event that one moves to another package, the other might as well and does not disrupt the call. Alternatively, we could just increase the visibility of `RemoteInvocationThread` to `public`, allowing the use of `instanceof`.

### DIFF
--- a/concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
+++ b/concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
@@ -53,6 +53,7 @@ import org.apache.thrift.transport.TServerSocket;
 import org.apache.thrift.transport.TTransportException;
 import org.cliffc.high_scale_lib.NonBlockingHashMap;
 
+import com.cinchapi.common.reflect.Reflection;
 import com.cinchapi.concourse.Constants;
 import com.cinchapi.concourse.Link;
 import com.cinchapi.concourse.Timestamp;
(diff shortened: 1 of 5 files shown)


concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -138,6 +140,11 @@
  * @author Jeff Nelson
  */
 public class ConcourseServer implements ConcourseService.Iface, ConcourseServerMXBean {
+    
+    /**
+     * The class representation of {@link RemoteInvocationThread}.
+     */
+    private static final Class<?> INVOCATION_THREAD_CLASS = Reflection.getClassCasted("com.cinchapi.concourse.server.plugin.RemoteInvocationThread");
```
jtnelson: Can you put this in GlobalState as a @NonPreference

jtnelson:
Lets keep ResultSet unparameterized and just make it use TObject instead of Object. Can the VisualizationEngine be switched to using TObject instead of Object. Logically, everything should work the same. At the very least, you can always convert a TObject to an Object.

We'd just need to edit the `getDatatTypeForClass` method in Dataset.java to handle TObject's appropriately by returning a DataType based on the TObject's type.

PR #210, opened 2016-08-31
CON-492-493

kylycht:
Includes [CON-492] [CON-493]


jtnelson:
Can you do a git checkout on your changes to ComplexTObject and then just add back the changes you made. It's showing that the whole file has been diffed which probably means you reformatted or something

concourse-server/src/main/java/com/cinchapi/concourse/server/plugin/PluginManager.java
```diff
@@ -458,6 +465,37 @@ private void initRealTimeStream(String id) {
...
 
     /**
+     * Create a {@link Packet} and place most recent {@link WriteEvent} on this
+     * {@link Packet}. Serialize {@link Packet} and place it on each of the
+     * {@link PluginManager} registered streams
+     *
+     * This starts to stream in separate thread
+     */
+    private void stream() {
+        final boolean isRunning = true;
```
jtnelson: this needs to be an instance variable
also call it `running` instead of `isRunning`

concourse-server/src/main/java/com/cinchapi/concourse/server/plugin/PluginManager.java
```diff
@@ -458,6 +465,37 @@ private void initRealTimeStream(String id) {
...
+        final ExecutorService executorService = Executors.newCachedThreadPool();
```
jtnelson: This should be an instance variable

jtnelson: Also, use `Executors.newFixedThreadPool` instead of a cached thread pool. You should pass in `Runtime.getRuntime().getAvailableProcessors()` as the parameter

concourse-server/src/main/java/com/cinchapi/concourse/server/plugin/PluginManager.java
```diff
@@ -458,6 +465,37 @@ private void initRealTimeStream(String id) {
...
+            @Override
+            public void run() {
+                while (isRunning) {
+                    List<WriteEvent> writeEvents = Lists.newArrayList();
+                    Queues.blockingDrain(BINARY_QUEUE, writeEvents);
+                    final Packet packet = new Packet(writeEvents);
+                    for (final SharedMemory stream : streams) {
+                        executorService.execute(new Runnable() {
+                            @Override public void run() {
+                                stream.write(Serializables.getBytes(packet));
```
jtnelson: You should get the bytes for the Packet once and just send that same serialized version to each of the packets

concourse-server/src/main/java/com/cinchapi/concourse/server/plugin/PluginManager.java
```diff
@@ -458,6 +465,37 @@ private void initRealTimeStream(String id) {
     }
 
     /**
+     * Create a {@link Packet} and place most recent {@link WriteEvent} on this
+     * {@link Packet}. Serialize {@link Packet} and place it on each of the
+     * {@link PluginManager} registered streams
+     *
+     * This starts to stream in separate thread
+     */
+    private void stream() {
```
jtnelson: This method will never be called outside of the context of the start routine, so you can add this logic directly to that method.

PR #220, opened 2016-09-16
CON-515 : cache clone dir, update clone if it still exists

kylycht:


concourse-ete-test-core/src/main/java/com/cinchapi/concourse/util/ConcourseCodebase.java
```diff
@@ -87,27 +87,51 @@ public static ConcourseCodebase cloneFromGithub() {
...
+                // If last cloned dir still exists use that clone, but perform
+                // git pull to pull latest changes
+                if(tmpDir != null) {
+                    StringBuilder sb = new StringBuilder();
+                    sb.append("git pull");
+                    try {
+                        LOGGER.info(
+                                "Running {} to pull latest changes from Github...",
+                                sb.toString());
+                        Process p = Runtime.getRuntime().exec(sb.toString());
```
jtnelson: This is going to run from the working directory of the java process and not from the directory where the code is. You'll need to use a `ProcessBuilder` which allows you to set the directory from which the process is launched

concourse-ete-test-core/src/main/java/com/cinchapi/concourse/util/ConcourseCodebase.java
```diff
@@ -162,6 +186,11 @@ private static String getTempDirectory() {
     private final String path;
 
     /**
+     * Reference to temp dir of last clone
+     * */
+    private static String tmpDir;
```
jtnelson: You don't need to store the tmpDir as an instance variable. Instead, you should store the path to the file in the user's home directory that is going to contain the location of the last used temporary directory

concourse-ete-test-core/src/main/java/com/cinchapi/concourse/util/ConcourseCodebase.java
```diff
@@ -87,27 +87,51 @@ public static ConcourseCodebase cloneFromGithub() {
...
-                            "Running {} to clone the concourse repo from Github...",
-                            sb.toString());
-                    Process p = Runtime.getRuntime().exec(sb.toString());
-                    int exitVal = p.waitFor();
-                    if(exitVal != 0) {
-                        throw new RuntimeException(Processes.getStdErr(p)
-                                .toString());
+                // If last cloned dir still exists use that clone, but perform
+                // git pull to pull latest changes
+                if(tmpDir != null) {
```
jtnelson: What I would do here is check to see if the user has the file in their home directory which contains the location of the last used temporary directory. If they do, read that location and see if it still exists. If it exists, launch a process that does git pull. If that process is successful, then set `dir` = to that location. Otherwise, go through the codepath of getting an new temp directory and re-cloning

concourse-ete-test-core/src/main/java/com/cinchapi/concourse/util/ConcourseCodebase.java
```diff
@@ -87,35 +94,77 @@ public static ConcourseCodebase cloneFromGithub() {
...
             }
             INSTANCE = new ConcourseCodebase(dir);
         }
         return INSTANCE;
     }
 
     /**
+     * File in user.home directory that will hold path to last clone
+     * */
+    private static final String clone_path = ".clone_path";
```
jtnelson: Please use java conventions and have camel case variable names

jtnelson: Make the relative path `.concourse-codebase`

concourse-ete-test-core/src/main/java/com/cinchapi/concourse/util/ConcourseCodebase.java
```diff
@@ -87,35 +94,77 @@ public static ConcourseCodebase cloneFromGithub() {
...
-                // and clone to some temp directory
-                dir = getTempDirectory();
-                StringBuilder sb = new StringBuilder();
-                sb.append("git clone ");
-                sb.append(GITHUB_CLONE_URL);
-                sb.append(" ");
-                sb.append(dir);
+                // If last cloned dir still exists use that clone, but perform
+                // git pull to pull latest changes
+                String tmpDir = null;
```
jtnelson: I don't think you need an extra variable, you can just use `dir`, since you'll ultimately need to set tmpDir = to dir anyway.

jtnelson:
@kylycht can you address the remaining code review feedback so we can merge this today.

concourse-ete-test-core/src/main/java/com/cinchapi/concourse/util/ConcourseCodebase.java
kylycht: updated

kylycht: updated

jtnelson:
merged manually

PR #222, opened 2016-09-16
(CON-517) Add logging to capture why unit test failed in TrackingMultimapVariableTypeTest#testNominal

as577:


concourse-plugin-core/src/test/java/com/cinchapi/concourse/plugin/data/TrackingMultimapVariableTypeTest.java
```diff
@@ -58,8 +59,13 @@ public void testDichotomous() {
     @Test
     public void testNominal() {
         int count = new java.util.Random().nextInt(10) + 3;  // in range (3, 12)
+        Variables.register("count", count);
         for(int i = 0; i < count; i++) {
-            map.insert(Random.getObject(), Random.getLong());
+            Object key = Random.getObject();
+            Long value = Random.getLong();
+            Variables.register("key", key);
```
jtnelson: These will get overwritten on each iteration of the loop, so you'll need to use keys like `"key_"+i` and `"value_"+i`. Also, you can use it when you create the object like
```
Object key = Variables.register("key_"+i, Random.getObject());

PR #228, opened 2016-09-21
updated vagrant file for ubuntu.

RaghavBabu:
upgraded to java8
memory set to 1024
fix made for vagrant.


jtnelson:
@RaghavBabu can you separate pull requests for the 1) virtualbox config changes and 2) the key insert? For each of those, make sure that the changes are done to all the Vagrantfiles

PR #231, opened 2016-09-25
Fixed private key issue bug in playground vagrant file.

RaghavBabu:


jtnelson:
@RaghavBabu all the Vagrantfiles need to have this change in this PR

PR #233, opened 2016-09-25
Configured memory for vm in vagrant file for centos and ubuntu.

RaghavBabu:


jtnelson:
What about playground?

PR #237, opened 2016-09-28
Added the ability to launch Concourse plugins in debug mode.

jonathanmarvens:


jtnelson:
How about we just have the `DEBUG_PORT` preference. The default port of 0 would imply the user does not want to run the plugin in DEBUG mode. Otherwise, if the port is set, that'll turn debug mode on. That way, the user only needs to flip one preference instead of 2.

jtnelson:
OR, we could have the `DEBUG_MODE` preference and choose a random port. The port can be printed in the plugin's logs so the user could find it and connect

concourse-plugin-core/src/main/java/com/cinchapi/concourse/server/plugin/PluginConfiguration.java
```diff
@@ -286,8 +286,8 @@ protected void addDefault(SystemPreference key, Object value) {
         LOG_LEVEL(null, String.class),
         ALIAS(null, ArrayList.class),
         ALIASES(null, ArrayList.class),
-        DEBUG_MODE(null, boolean.class, Boolean.class),
-        DEBUG_PORT(null, int.class, Integer.class);
+        REMOTE_DEBUGGER(null, boolean.class, Boolean.class),
```
jonathanmarvens: __Example:__ `remote_debugger = ON` in __`$PLUGIN_HOME/conf/plugin.prefs.dev`__.

concourse-plugin-core/src/main/java/com/cinchapi/concourse/server/plugin/PluginConfiguration.java
```diff
@@ -286,8 +286,8 @@ protected void addDefault(SystemPreference key, Object value) {
...
+        REMOTE_DEBUGGER_PORT(null, int.class, Integer.class);
```
jonathanmarvens: __Example:__ `remote_debugger_port = 48410` in __`$PLUGIN_HOME/conf/plugin.prefs.dev`__.

jonathanmarvens: __Example:__

In `$PLUGIN_HOME/conf/plugin.prefs.dev`:

```
# Default: OFF
remote_debugger = ON
# Default: 48410
remote_debugger_port = 26342
```

PR #240, opened 2016-09-30
Added feature to get and set values from concourse.prefs. Handled bou…

RaghavBabu:
…ndary cases.


concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -3954,6 +3961,12 @@ public void setKeyValueRecords(String key, TObject value,
     }
 
     @Override
+    public void setPreference(String key, String value) {
```
RaghavBabu: In ConcourseMXBean, I couldn't use object.. I get this below exception.
com.cinchapi.concourse.server.jmx.ConcourseServerMXBean.setPreference has parameter or return type that cannot be translated into an open type.
So I made it string.. because we ultimately going to print a string in console.

PR #242, opened 2016-10-03
CON-513: Added logic to shutdown JavaApp on host process abnormal termination

RaghavBabu:


concourse-driver-java/src/main/java/com/cinchapi/concourse/util/Processes.java
```diff
@@ -140,6 +144,68 @@ public static void waitForSuccessfulCompletion(Process process) {
         return readStream(process.getErrorStream());
     }
 
+    /**
+     * Get current process string representation of process id.
+     * 
+     * @param process
+     * @return String representation of process id
+     */
+    public static String getCurrentPid(Process process) {
```
jtnelson: Why are you passing a `process` in that is not getting used?

concourse-driver-java/src/main/java/com/cinchapi/concourse/util/Processes.java
```diff
@@ -140,6 +144,68 @@ public static void waitForSuccessfulCompletion(Process process) {
...
+            else if(SystemUtils.IS_OS_WINDOWS) {
+                process = Runtime.getRuntime().exec(
+                        "TASKLIST /fi \"PID eq " + pid + "\" /fo csv /nh");
+            }
+        }
+        catch (Exception e) {
+            Throwables.propagate(e);
+        }
+        if(process != null) {
+            BufferedReader in = new BufferedReader(
```
jtnelson: use the `readStream` method

concourse-driver-java/src/main/java/com/cinchapi/concourse/util/Processes.java
```diff
@@ -140,6 +144,68 @@ public static void waitForSuccessfulCompletion(Process process) {
...
+                }
+            }
+            catch (IOException e) {
+                Throwables.propagate(e);
+            }
+        }
+        else {
+            return false;
+        }
+        return true; // JavaApp should not shutdown because of exception while
```
jtnelson: Don't make a reference to `JavaApp` here. The `Processes` class is general and not used only for JavaApps.

PR #243, opened 2016-10-03
CON-528 

kylycht:


concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -137,25 +139,42 @@
...
+            TServerTransport serverTransport = null;
+            try {
+                serverTransport = new TServerSocket(JMX_PORT);
+            }
+            catch (TTransportException e) {
+                e.printStackTrace();
+            }
+            TServer managementServer = new TSimpleServer(
+                    new Args(serverTransport).processor(processor));
+            managementServer.serve();
```
jtnelson: You should call `serve` on the managementServer in the ConcourseServer's `start` method

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -137,25 +139,42 @@
...
+    private void createManagementServer() throws TTransportException{
+        ConcourseManagementService.Processor<ConcourseManagementService.Iface>
+                processor = new ConcourseManagementService.Processor<>(this);
+        Runnable managementSimple = () -> {
+            TServerTransport serverTransport = null;
+            try {
+                serverTransport = new TServerSocket(JMX_PORT);
+            }
+            catch (TTransportException e) {
+                e.printStackTrace();
```
jtnelson: never do `e.printStackTrace` use `throw Throwables.propagate(e)` instead

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -137,25 +139,42 @@
...
+     *
      * @return {@link ConcourseServer}
      * @throws TTransportException
      */
     public static ConcourseServer create() throws TTransportException {
         return create(CLIENT_PORT, BUFFER_DIRECTORY, DATABASE_DIRECTORY);
     }
 
+    private void createManagementServer() throws TTransportException{
+        ConcourseManagementService.Processor<ConcourseManagementService.Iface>
```
jtnelson: In general, i think you should only create the managementServer instance in the `init` method of ConcourseServer. In Concourse Server's `start` method you should have a daemon thread that calls serve on the managementServer.

jtnelson:
@kylycht can you also add logic to `utils/compile-thrift-java.sh` to generate the thrift sources for the management code?

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -2621,29 +2688,12 @@ public String getServerEnvironment(AccessToken creds,
     }
 
     @Override
-    @ManagedOperation
-    public String getServerVersion() {
+    public String getServerVersion()
+            throws SecurityException, TransactionException, TException {
```
jtnelson: why does this need to declare that it throws these exceptions, now?

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -3986,6 +3976,11 @@ public void start() throws TTransportException {
         plugins.start();
         System.out.println("The Concourse server has started");
         server.serve();
+        Thread managementd = new Thread(() -> {
```
jtnelson: You have to run this thread before the call to `sever.serve()` because that method is blocking.

utils/compile-thrift-java.sh
```diff
@@ -53,4 +53,11 @@ groovy $GENERATOR $THRIFT_IDL $SOURCE_DESTINATION
 
 echo "Finished generating $SOURCE_DESTINATION"
 
+#Generate the ConcourseManagementService class
+THRIFT_IDL=$HOME/../interface/management/management.thrift
+SOURCE_DESTINATION=$HOME/../concourse-driver-java/src/main/java
```
jtnelson: these sources should be placed in the `concourse-server` project instead of `concourse-driver-java`

jtnelson:
Tracking at https://github.com/cinchapi/concourse/pull/245

PR #244, opened 2016-10-04
Added logic to shutdown JavaApp on host process abnormal termination

RaghavBabu:


concourse-server/src/main/java/com/cinchapi/concourse/server/io/process/JavaApp.java
```diff
@@ -170,6 +182,30 @@ public JavaApp(String classpath, String source, String... options) {
         this.sourceFile = workspace + separator + mainClass + ".java";
         FileSystem.write(source, sourceFile);
 
+        // Thread which phones host process and checks its status for every 5
+        // seconds. Terminates itself, if host process is down.
+        String processId = ManagementFactory.getRuntimeMXBean().getName()
```
jtnelson: Can you put this logic in a util class. There is a class called `Processes` in accent4j where you can add a method called `getCurrentPid`

concourse-server/src/main/java/com/cinchapi/concourse/server/io/process/ProcessValidator.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * ProcessValidator provides functionality to check if particular process id is
+ * running and returns true if it is else it return false. It is designed to be
+ * platform-independent. It works for both Linux and Windows platform.
+ * 
+ * @author raghavbabu
+ *
+ */
+public class ProcessValidator {
```
jtnelson: I think these methods would be better placed in the `Processes` util class as opposed to creating another util class.

concourse-server/src/main/java/com/cinchapi/concourse/server/io/process/ProcessValidator.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Check if the process with the processId is running.
+     * 
+     * @param processId Id for the input process.
+     * @return true if its running, false if not.
+     */
+    protected static boolean isProcessRunning(String processId) {
```
jtnelson: use variable name `pid` instead of `processId`

concourse-server/src/main/java/com/cinchapi/concourse/server/io/process/ProcessValidator.java
```diff
@@ -0,0 +1,72 @@
...
+
+        Process process = null;
+        try {
+            if(SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_MAC
+                    || SystemUtils.IS_OS_MAC_OSX || SystemUtils.IS_OS_UNIX
+                    || SystemUtils.IS_OS_SOLARIS || SystemUtils.IS_OS_SUN_OS) {
+                process = Runtime.getRuntime().exec("pgrep -f java");
```
jtnelson: I don't think `pgrep` is installed by default on most systems. I think it'd be safer to use `ps aux | grep <pid>`. In order to do that you'll need to create a process builder with pipe support using `com.cinchapi.concourse.util.Processes#getBuilderWithPipeSupport`

concourse-server/src/main/java/com/cinchapi/concourse/server/io/process/ProcessValidator.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Parse through the process input stream to check if it contains the
+     * specific process id.
+     * 
+     * @param processId
+     * @param process
+     * @return true if process id present else returns false;
+     */
+    private static boolean parseOutput(String processId, Process process) {
```
jtnelson: I don't think this logic needs to be in a separate method because it is very specific and unlikely to ever be called from any other function.

PR #247, opened 2016-10-07
Renamed cli classes and scripts

RaghavBabu:


concourse-server/scripts/concourse
```diff
@@ -180,7 +180,7 @@ JVMOPTS=$JVMOPTS"
 #
 # The format for this array is alias:cli
 BIN_ALIASES=(
-  debug:dumptool
```
jtnelson: Can you keep `debug` as an alias for dumptool. I want both `datafiles` and `debug` to be aliases

PR #249, opened 2016-10-08
CON-537: Fix size() call in ObjectResultDataset

as577:


concourse-server/log/console.log
```diff
@@ -0,0 +1,2 @@
+Error: Could not find or load main class com.cinchapi.concourse.server.ConcourseServer
```
jtnelson: can you uncommit this file?

PR #252, opened 2016-10-16
Added util method to Processes

RaghavBabu:


concourse-driver-java/src/main/java/com/cinchapi/concourse/util/Processes.java
```diff
@@ -129,7 +130,17 @@ public static void waitForSuccessfulCompletion(Process process) {
             throw Throwables.propagate(e);
         }
     }
-
+    
+    /**
+     * Get current process string representation of process id.
+     * 
+     * @param process
```
jtnelson: fix javadoc. There is no param to this method

concourse-driver-java/src/main/java/com/cinchapi/concourse/util/Processes.java
```diff
@@ -129,7 +130,16 @@ public static void waitForSuccessfulCompletion(Process process) {
             throw Throwables.propagate(e);
         }
     }
-
+    
+    /**
+     * Get current process string representation of process id.
```
jtnelson: `Return the pid of the current process`

PR #253, opened 2016-10-18
CON-502: A -ps flag to plugin CLI

RaghavBabu:
Added plugin status to ManagePluginCli.. Displays info of all currently running plugins.


jtnelson:
@RaghavBabu use a UUID for the app token

PR #258, opened 2016-11-07
CON-556: System Id

RaghavBabu:


concourse-server/src/main/java/com/cinchapi/concourse/server/GlobalState.java
```diff
@@ -85,6 +86,13 @@
      */
     public static String BUFFER_DIRECTORY = System.getProperty("user.home")
             + File.separator + "concourse" + File.separator + "buffer";
+    
+    /**
+     * This {@link UUID} is used to identify the Concourse instance across host and port
+     * changes. This id will be registered locally in the system in data
+     * and buffer directory.
+     */
```
jtnelson: Add `@NonPreference` annotation

jtnelson: Why are you setting this to a random UUID? That means it'll be different every time the server starts

jtnelson:
1. The logic for checking the system id should live in GlobalState. There you should simply check the id from the BUFFER and DATABASE directories. If they are the same, set the SYSTEM_ID property in GlobalState. If they are not the same, set the property to `null`.
2. In Concourse Server, check that GlobalState.SYSTEM_ID is not null. If it is, throw an error.

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -208,6 +209,13 @@ public static void main(String... args) throws TTransportException,
 
         // Create an instance of the server and all of its dependencies
         final ConcourseServer server = ConcourseServer.create();
+        GlobalState.compareAndsetSystemId();
+        // Check if concourse is in inconsistent state.
+        if(GlobalState.SYSTEM_UUID.equals(null)) {
```
jtnelson: use `==`

concourse-server/src/main/java/com/cinchapi/concourse/server/GlobalState.java
```diff
@@ -87,6 +90,15 @@
             + File.separator + "concourse" + File.separator + "buffer";
 
     /**
+     * This {@link UUID} is used to identify the Concourse instance across host
+     * and port
+     * changes. This id will be registered locally in the system in data
+     * and buffer directory.
+     */
+    @NonPreference
```
jtnelson: add `@Nullable` annotation as well

concourse-server/src/main/java/com/cinchapi/concourse/server/GlobalState.java
```diff
@@ -87,6 +90,15 @@
             + File.separator + "concourse" + File.separator + "buffer";
 
     /**
+     * This {@link UUID} is used to identify the Concourse instance across host
+     * and port
```
jtnelson: fix formatting

concourse-server/src/main/java/com/cinchapi/concourse/server/GlobalState.java
```diff
@@ -429,4 +441,80 @@ public static String getPrefsFilePath() {
...
+     * inconsistency. If the file is present, it reads the file from both
+     * directories and return a boolean after comparison.
+     * 
+     * <p>
+     * Set the SYSTEM_ID if both the id are same in database and buffer
+     * directory otherwise it will set it to null.
+     * </p>
+     * 
+     */
+    public static void compareAndsetSystemId() {
```
jtnelson: method name should be properly camel cased

jtnelson: Instead of having this logic in a method that is invoked from Concourse Server, it should be called automatically in a static initialization block inside `GlobalState`

concourse-server/src/main/java/com/cinchapi/concourse/server/GlobalState.java
```diff
@@ -429,4 +441,80 @@ public static String getPrefsFilePath() {
...
+        boolean recent = false;
+        String bufferIdFile = BUFFER_DIRECTORY + File.separator + ".id";
+        String dbIdFile = DATABASE_DIRECTORY + File.separator + ".id";
+        UUID systemId = UUID.randomUUID();
+        String uuidString = systemId.toString();
+        StringBuilder builder = new StringBuilder();
+        for (int i = 0; i < uuidString.length(); i++) {
```
jtnelson: Use `ByteBuffers#fromString` instead of manually doing this

concourse-server/src/main/java/com/cinchapi/concourse/server/GlobalState.java
```diff
@@ -429,4 +441,80 @@ public static String getPrefsFilePath() {
...
+        // write to database .id file if file not present.
+        ByteBuffer dbId = null;
+        if(!FileSystem.hasFile(dbIdFile)) {
+            FileSystem.writeBytes(uuidBuffer, dbIdFile);
+        }
+        else {
+            dbId = FileSystem.readBytes(dbIdFile);
+        }
+        if(recent) {
+            System.out.println("recent");
```
jtnelson: remove this

jtnelson:
@RaghavBabu there's still some feedback to address on this PR

PR #265, opened 2016-11-22
CON-292: Link navigation

RaghavBabu:


concourse-driver-java/src/main/java/com/cinchapi/concourse/Concourse.java
```diff
@@ -415,6 +415,298 @@ public Object calculate(String method, Object... args) {
...
+    /**
+     * Return all the values stored for {@code key} in {@code record}.Iterates
+     * through the key splitted with dot(.) operator. Navigates only if the key
+     * has a link as value which points to another record.
+     * 
+     * @param key the field name
+     * @param record the record id
+     * @return a {@link Set} containing all the values stored in the field
+     */
+    public abstract <T> Set<T> navigate(String key, long record);
```
jtnelson: this has to return a `Map<Long, Set<T>>`

concourse-integration-tests/src/test/java/com/cinchapi/concourse/NavigationTest.java
```diff
@@ -0,0 +1,415 @@
...
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import com.beust.jcommander.internal.Lists;
```
jtnelson: you should be importing `Lists` and `Sets` from the guava packages instead of jcommander

PR #267, opened 2016-11-25
CON-563: Add "max" aggregation function

RaghavBabu:


jtnelson:
@RaghavBabuatomic.browse(key) will return the values in sorted order, so for maxKey you can simply return the last value in the keyset.

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -848,6 +848,420 @@ public TObject averageKeyTime(String key, long timestamp, AccessToken creds,
...
+            throws TException {
+        checkAccess(creds, transaction);
+        AtomicSupport store = getStore(transaction, environment);
+        AtomicOperation atomic = null;
+        TObject max = null;
+        while (atomic == null || !atomic.commit()) {
+            atomic = store.startAtomicOperation();
+            try {
+                Map<TObject, Set<Long>> data = atomic.browse(key);
+                int size = data.size();
```
jtnelson: Use `Iterables.getLast`

PR #268, opened 2016-11-26
CON-562: Add "min" aggregation function

RaghavBabu:


jtnelson:
@RaghavBabu`atomic.browse(key)` will return the values in sorted order, so for `minKey` you can simply return the last value in the keyset.

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -848,6 +848,403 @@ public TObject averageKeyTime(String key, long timestamp, AccessToken creds,
...
+            throws TException {
+        checkAccess(creds, transaction);
+        AtomicSupport store = getStore(transaction, environment);
+        AtomicOperation atomic = null;
+        TObject min = null;
+        while (atomic == null || !atomic.commit()) {
+            atomic = store.startAtomicOperation();
+            try {
+                Map<TObject, Set<Long>> data = atomic.browse(key);
+                for (Entry<TObject, Set<Long>> entry : data.entrySet()) {
```
jtnelson: Use `Iterables.getFirst`

PR #287, opened 2017-06-13
feature(Add describe api's)

aunyks:
Adding `describe`, `describeTime`, and `describeTimeStr` methods.


aunyks:
PR no longer a WIP

interface/concourse.thrift
```diff
@@ -1102,6 +1102,84 @@ service ConcourseService {
...
+   * @param timestamp the historical timestamp to use in the lookup
+   * @return the {@link Set} of keys that were in the database at
+   *         {@code timestamp}
+   * @throws exceptions.SecurityException if the {@code creds} don't
+   *         represent a valid session
+   * @throws exceptions.TransactionException if the client was in a
+   *         transaction and an error occurred that caused the transaction
+   *         to end itself
+   */
+  set<string> describeTime(
```
jtnelson: @aunyks the `string` and `timestamp` parameters should come before the `creds`

interface/concourse.thrift
```diff
@@ -1102,6 +1102,84 @@ service ConcourseService {
...
+   * @param timestamp the historical timestamp to use in the lookup
+   * @return the {@link Set} of keys that were in the database at
+   *         {@code timestamp}
+   * @throws exceptions.SecurityException if the {@code creds} don't
+   *         represent a valid session
+   * @throws exceptions.TransactionException if the client was in a
+   *         transaction and an error occurred that caused the transaction
+   *         to end itself
+   */
+  set<string> describeTimeStr(
```
jtnelson: @aunyks see other comment about order of params

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -1152,6 +1152,80 @@ public boolean commit(AccessToken creds, TransactionToken transaction,
...
+                atomic = null;
+            }
+        }
+        return result;
+    }
+    
+    @Atomic
+    @HistoricalRead
+    @ThrowsThriftExceptions
+    public Set<String> describeTime(AccessToken creds, TransactionToken transaction,
```
jtnelson: @aunyks the name of this method should be `describeTimestr` to be consistent with the thrift api

interface/concourse.thrift
jtnelson: call this method `describeTimestr` for consistency with the other method names that have `Timestr` instead of `TimeStr`

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -1152,6 +1152,80 @@ public boolean commit(AccessToken creds, TransactionToken transaction,
...
+            String environment, String timestamp) throws TException {
+        checkAccess(creds, transaction);
+        AtomicSupport store = getStore(transaction, environment);
+        Set<String> result = Sets.newHashSet();
+        Timestamp stamp = Timestamp.parse(timestamp,
```
jtnelson: @aunyks ConcourseServer use a unix timestamp instead of a `Timestamp` object so here you want to use `NaturalLanguge.parseMicros(timestamp) to get the micros that you'll pass to the `store.describe` method.

concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
jtnelson: @aunyks for `describeTimestr` you can actually implement it like this
```java
public Set<String> describeTimestr(String timestamp, AccessToken creds, TransactionToken transaction, String environment) {
    return describeTime(NaturalLanguage.parseMicros(timestamp), creds, transaction, environment);
}
```
In this instance it is okay to call another server method (instead of call into the engine) because you haven't done any work that would be repeated by the method you're calling.

aunyks:
@jtnelson How's it look now?

jtnelson:
@aunyks So far this is looking great!

Next step is to implement these new methods in `ConcourseThriftDriver.java`. You'll want to create the following overloaded method signatures:
`describe()`
`describe(Timestamp timestamp)`

and in each of those call the appropriate Thrift method using the `client` object. Check out the other implementations in that file for information about how to wrap all of this in an `execute` call.

aunyks:
@jtnelson For some reason, even after adding the new abstract `describe()` method signatures to `Concourse.java` they're still now acknowledged in the codebase.

In saying that I mean when I added the new methods to `Concourse.java` they don't show in CaSH. Only the old two do.

concourse-integration-tests/src/test/java/com/cinchapi/concourse/DescribeTest.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public class DescribeTest extends ConcourseIntegrationTest {
+	
+	@Test
+	public void testDescribeNoTimestamp(){
+		client.add("name", "Gerald");
+		client.add("address", "123 Main St");
+		client.add("title", "Person");
+		
+		String[] expecteds = {"address" , "name", "title"};
```
jtnelson: @aunyks you should make `expected` a Set to begin with so you don't have to convert the return value of the `describe` method to an array

concourse-integration-tests/src/test/java/com/cinchapi/concourse/DescribeTest.java
```diff
@@ -0,0 +1,58 @@
...
+	}
+	
+	@Test
+	public void testDescribeWithTimestamp(){
+		client.add("name", "Gerald");
+		client.add("address", "123 Main St");
+		Timestamp stamp = Timestamp.now();
+		client.add("title", "Person");
+		
+		String[] expecteds = {"address", "name"};
```
jtnelson: @aunyks you should make expected a Set to begin with so you don't have to convert the return value of the describe method to an array

aunyks:
@jtnelson Look a bit better?

PR #300, opened 2017-11-20
CON-593: Provide more information when plugin errors occur.

jtnelson:


jtnelson:
Testing this requires testing a plugin that depends on this updated version of the code.

PR #301, opened 2017-11-20
Add better error message for TApplicationException in CaSH

jtnelson:


jtnelson:
I need to update the CHANGELOG

PR #310, opened 2017-12-03
CON-601: Remove CCL parsing code from concourse in favor of external CCL parsing library

jtnelson:


jtnelson:
This PR is code complete. Unit tests are passing. Do some more manual testing and then merge.

PR #323, opened 2018-01-07
Upgrade to CCL 2.3.0

jtnelson:


jtnelson:
Still waiting on fix for CCL-7 and for the official CCL 2.3.0 release so we're not using a snapshot version.

PR #364, opened 2019-05-09
CON-639: Order filter logic

javierlores:


concourse-server/src/main/java/com/cinchapi/concourse/server/ConcourseServer.java
```diff
@@ -4258,6 +4260,73 @@ public void revertKeysRecordTimestr(List<String> keys, long record,
...
+
+        return Sorter.sort(result, order);
+    }
+
+    @Override
+    @ThrowsClientExceptions
+    @VerifyAccessToken
+    @VerifyReadPermission
+    public Map<Long, Map<String, Set<TObject>>> selectKeysCriteriaTimeOrder(
+            List<String> keys, TCriteria criteria, long timestamp, Order order,
```
jtnelson: @javierlores How is `Order` being represented within the thrift layer?

javierlores: Yeah, currently working on adding that

PR #370, opened 2019-06-21
adjusted flag value as result of JEP-245

robertcli:
Fix for higher version of Java that correct JVM flag "hack", JEP-245.


jtnelson:
@asyncRobert I think I mentioned how this needs to be tested, on the Jira ticket, correct?

asyncRobert:
@jtnelson Yes! You mentioned that on Jira. I will test this and get back with results.

jtnelson:
Merged in https://github.com/cinchapi/concourse/pull/387

PR #371, opened 2019-06-21
Add remote debugger port option to concourse.prefs

asyncRobert:
CON-577

Add remote debugger port option to Concourse prefs.

### DIFF
--- a/concourse-server/conf/concourse.prefs
+++ b/concourse-server/conf/concourse.prefs
@@ -148,3 +148,7 @@
 #
 # DEFAULT: automatically chosen based on the client_port
 #shutdown_port = 3434
+
+# The remote debugger port.
+# DEFAULT: 54555
+#remote_debugger_port = 54555
--- a/concourse-server/conf/concourse.prefs
+++ b/concourse-server/conf/concourse.prefs
@@ -153,5 +153,6 @@
 # 49152 and 65535 to minimize the possibility of conflicts with other services
 # on this host. If the value of this preference is set to 0, then remote debugging for
 # Concourse Server is disabled.
+#
 # DEFAULT: 0
 #remote_debugger_port = 0
--- a/concourse-server/conf/concourse.prefs
+++ b/concourse-server/conf/concourse.prefs
@@ -149,6 +149,9 @@
 # DEFAULT: automatically chosen based on the client_port
 #shutdown_port = 3434
 
-# The remote debugger port.
-# DEFAULT: 54555
-#remote_debugger_port = 54555
+# The listener port (1-65535) for remote debugger connections. Choose a port between
+# 49152 and 65535 to minimize the possibility of conflicts with other services
+# on this host. If the value of this preference is set to 0, then remote debugging for
+# Concourse Server is disabled.
+# DEFAULT: 0
+#remote_debugger_port = 0
--- a/concourse-server/scripts/concourse
+++ b/concourse-server/scripts/concourse
@@ -233,6 +233,13 @@ if [ -z "$JMX_PREF" ]; then
 fi
 JVMOPTS=${JVMOPTS/jmxremote.port=9010/jmxremote.port=$JMX_PREF}
 
+#remote_debugger_port
+RDP_PREF=`cat $PREFS | grep -e '^remote_debugger_port\s*=\s*[0-9]\{1,\}$' | head -n1 | cut -d'=' -f2 | tr -d ' '`
+if [ -z "$RDP_PREF" ]; then
+  RDP_PREF="54555"
+fi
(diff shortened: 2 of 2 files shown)


concourse-server/scripts/concourse
```diff
@@ -57,7 +57,7 @@ JVMOPTS="
 -Dcom.sun.management.jmxremote.authenticate=false
 -Dcom.sun.management.jmxremote.ssl=false
 -XX:+UseThreadPriorities
--XX:ThreadPriorityPolicy=42
+-XX:ThreadPriorityPolicy=1
```
jtnelson: @asyncRobert Can you remove this change from this branch since it relates to another ticket.

asyncRobert: Oops, reverted.

concourse-server/conf/concourse.prefs
```diff
@@ -148,3 +148,7 @@
 #
 # DEFAULT: automatically chosen based on the client_port
 #shutdown_port = 3434
+
+# The remote debugger port.
```
jtnelson: @asyncRobert Set this as the pref description

```
# The listener port (1-65535) for remote debugger connections. Choose a port between
# 49152 and 65535 to minimize the possibility of conflicts with other services
# on this host. If the value of this preference is set to 0, then remote debugging for
# Concourse Server is disabled.
```

concourse-server/conf/concourse.prefs
```diff
@@ -148,3 +148,7 @@
...
+# DEFAULT: 54555
```
jtnelson: @asyncRobert Set the default to `0`

concourse-server/scripts/concourse
```diff
@@ -233,6 +233,12 @@ if [ -z "$JMX_PREF" ]; then
 fi
 JVMOPTS=${JVMOPTS/jmxremote.port=9010/jmxremote.port=$JMX_PREF}
 
+#remote_debugger_port
+RDP_PREF=`cat $PREFS | grep -e '^remote_debugger_port\s*=\s*[0-9]\{1,\}$' | head -n1 | cut -d'=' -f2 | tr -d ' '`
+if [ -n "$RDP_PREF" ]; then
```
jtnelson: @asyncRobert Add a check to verify that the RDP_PREF value is not equal to `0`. If the pref is equal to 0, you should treat it as if the RDP_PREF is not provided.

concourse-server/scripts/concourse
```diff
@@ -233,6 +233,12 @@ if [ -z "$JMX_PREF" ]; then
...
+  JVMOPTS=${JVMOPTS/-Xdebug/-Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=$RDP_PREF}
```
jtnelson: @asyncRobert This logic is actually doing a find/replace which won't work in this case (sorry if I misled you in the original ticket)

You actually want to add the remote debugging flags to the JVMOPTS by doing
```
if [ -n "$RDP_PREF" ]; then
  JVMOPTS=$JVMOPTS"
  -Xdebug
  -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=$RDP_PREF
  "
fi
```

concourse-server/scripts/concourse
```diff
@@ -233,6 +233,15 @@ if [ -z "$JMX_PREF" ]; then
 fi
 JVMOPTS=${JVMOPTS/jmxremote.port=9010/jmxremote.port=$JMX_PREF}
 
+#remote_debugger_port
+RDP_PREF=`cat $PREFS | grep -e '^remote_debugger_port\s*=\s*[0-9]\{1,\}$' | head -n1 | cut -d'=' -f2 | tr -d ' '`
+if [ "$RDP_PREF" -ne "0" ]; then
```
jtnelson: @asyncRobert What happens if RDP_PREF is empty. Would this check pass and the remote debugger flags would be added?

asyncRobert: I tested alternative logic that works with both scenarios and pushed to CON-577, let me know what you think!

concourse-server/scripts/concourse
```diff
@@ -233,6 +233,17 @@ if [ -z "$JMX_PREF" ]; then
 fi
 JVMOPTS=${JVMOPTS/jmxremote.port=9010/jmxremote.port=$JMX_PREF}
 
+#remote_debugger_port
+RDP_PREF=`cat $PREFS | grep -e '^remote_debugger_port\s*=\s*[0-9]\{1,\}$' | head -n1 | cut -d'=' -f2 | tr -d ' '`
+if [ ! -z "$RDP_PREF" ]; then
+     if [ "$RDP_PREF" -ne "0" ]; then
```
jtnelson: @asyncRobert you should write this as one conditional similar to https://stackoverflow.com/a/13408590.

asyncRobert: pushed one conditional statement

PR #374, opened 2019-06-25
CON-651: Support link navigation in selection keys

javierlores:
Depends on CON-573


concourse-server/src/main/java/com/cinchapi/concourse/server/query/Finder.java
```diff
@@ -99,6 +110,7 @@ private Finder() {/* singleton */}
         Verify.that(data.length >= 1);
         Verify.that(data[0] instanceof Store);
         Store store = (Store) data[0];
+        AtomicOperation atomic = (AtomicOperation) data[0];
```
jtnelson: @javierlores This isn't a safe cast because the `store` will not be an `AtomicOperation` for some usages.

Understanding that navigable select keys must be done within an AtomicOperation, I'd suggest first checking each of the keys in the ConcourseServer methods. If any of those keys are navigable, you should ensure, within ConcourseServer that the store is an AtomicOperation.

concourse-server/src/main/java/com/cinchapi/concourse/server/query/Finder.java
```diff
@@ -125,11 +137,38 @@ else if(operator == Operator.NOT_EQUALS) {
...
+            Set<Long> results;
+            if(LinkNavigation.isNavigationScheme(key)) {
+                TObject[] builtValues = values.build();
+                results = Sets.newHashSet();
+                Set<Long> records = store.getAllRecords();
+                key = expression.key().toString();
+
+                for (long record : records) {
+                    Map<Long, Set<TObject>> result = Operations
+                            .navigateKeyRecordAtomic(key, record, Time.NONE,
```
jtnelson: Shouldn't the timestamp that is passed in be that associated with the expression and not always `Time.NONE`? @javierlores

concourse-server/src/main/java/com/cinchapi/concourse/server/query/Finder.java
```diff
@@ -125,11 +137,38 @@ else if(operator == Operator.NOT_EQUALS) {
...
                     value -> values.add(Convert.javaToThrift(value.value())));
-            Set<Long> results = expression
-                    .timestamp() == TimestampSymbol.PRESENT
-                            ? store.find(key, operator, values.build())
-                            : store.find(expression.raw().timestamp(), key,
-                                    operator, values.build());
+
+            // If the key is a navigation key
+            Set<Long> results;
+            if(LinkNavigation.isNavigationScheme(key)) {
```
jtnelson: @javierlores checking all the records seems like a recipe for slow performance. I understand why you have to take that approach so i'd recommend investigating what it would take to support navigable find/selection within the Engine level. Let me know if you need guidance on this.

jtnelson: @javierlores Another option is to leverage the `browse` functionality provided by the `Store` which returns a map from each value stored under a key to the Set of records in which that value is contained. Going this route means that you couldn't use `Operations.navigateKeyRecordAtomic`. You probably need to create a new utility function that does something along the lines of

- break up the navigation key into parts
- for each part, perform a `browse` and filter away all the values in the returned maps `keySet` that are not links; you'll also need to collect all the mapped records (e.g. values().flatMap()) and on successive iterations also filter out records that are not in the running candidate set of records
- finally once you've gone through each part of the navigation key, check the last set of values returned from browse() and include the records for the values that satisfy the expression operation in relation to the expression values

concourse-server/src/main/java/com/cinchapi/concourse/server/query/Finder.java
javierlores: Yup, I forgot to change from when I first wrote it.

concourse-server/src/main/java/com/cinchapi/concourse/server/query/Finder.java
javierlores: Agreed, this was the implementation suggestion written in the Jira ticket. So I had already looked into supporting the navigable find/selection within the Engine level for the same reason and so I think I can figure it out, but stay tuned. I had not thought of the other option though. If performance is the main concern, I think implementing at the engine level would be ideal no?

javierlores: Lol, didn't see your last message saying the same thing. I'll try implementing at the engine level then

concourse-server/src/main/java/com/cinchapi/concourse/server/query/Finder.java
javierlores: You right, I'll fix.

PR #375, opened 2019-06-26
CON-625: Check for plugin restricted methods in Plugins that have a parent class

Tylerwbrown:
http://jira.cinchapi.com/browse/CON-625

### DIFF
--- a/concourse-plugin-core/src/main/java/com/cinchapi/concourse/server/plugin/RemoteInvocationThread.java
+++ b/concourse-plugin-core/src/main/java/com/cinchapi/concourse/server/plugin/RemoteInvocationThread.java
@@ -28,7 +28,6 @@ import com.cinchapi.concourse.thrift.AccessToken;
 import com.cinchapi.concourse.thrift.ComplexTObject;
 import com.cinchapi.concourse.thrift.TransactionToken;
 import com.google.common.base.Throwables;
-import static com.cinchapi.common.reflect.Reflection.isDeclaredAnnotationPresentInHierarchy;
 
 /**
  * A daemon {@link Thread} that is responsible for processing a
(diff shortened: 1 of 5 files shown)


concourse-plugin-core/src/main/java/com/cinchapi/concourse/server/plugin/RemoteInvocationThread.java
```diff
@@ -28,6 +28,7 @@
 import com.cinchapi.concourse.thrift.ComplexTObject;
 import com.cinchapi.concourse.thrift.TransactionToken;
 import com.google.common.base.Throwables;
+import static com.cinchapi.common.reflect.Reflection.isDeclaredAnnotationPresentInHierarchy;
```
jtnelson: @Tylerwbrown By convention, we don't do static imports in the codebase. Can you change it to call `Reflection.isDeclaredAnnotationPresentInHierarchy` below

jtnelson:
@Tylerwbrown You should also create a unit test for this in https://github.com/cinchapi/concourse/tree/develop/concourse-plugin-core-tests/src/test/java/com/cinchapi/concourse/server/plugin

To do so, you'll need to create a Plugin that extends https://github.com/cinchapi/concourse/blob/develop/concourse-plugin-core-tests/src/test/java/com/cinchapi/concourse/server/plugin/TestPlugin.java and annotate one of the methods in TestPlugin as @PluginRestricted (NOTE: You should create a new method with this annotation)

Then create a test class that `extends ClientServerTest implements PluginTest` and try to invoke the method that is PluginRestricted

concourse-plugin-core/src/main/java/com/cinchapi/concourse/server/plugin/RemoteInvocationThread.java
Tylerwbrown: @jtnelson Sure thing, is there any large downside to static imports that swayed us towards that idiom? Or did we choose to not do it and we're trying to keep it consistent? Curious for personal projects of mine as well :)

jtnelson: The pro of static imports is that it can make code more readable

The con of static imports is that it can make code less immediately understandable (i.e. thinking a method is defined in a class when its actually statically imported from another class).

We've never used them in the codebase and since the project is fairly large, I think the cost of the con outweighs the pro; especially since readability is pretty subjective

PR #376, opened 2019-06-29
upgrade to Gradle 4.10.2

jtnelson:


jtnelson:
There seems to be an issue with the spotless plugin not working properly and spinning forever.

PR #377, opened 2019-06-30
Don't track stats about result datasets in ConcourseServer

jtnelson:
Implement and use a `LazyTrackingResultDataset` in Concourse Server so starts are tracked prior to serialization but are tracked after deserialization in the receiving plugin/service.

The purpose of this change is to make sure that Concourse Server doesn't waste cycles tracking stats for plugins twice. Currently, ConcourseServer will track the stats when populating the result set and the plugin will track them again when deserializing the data it receives from ConcourseRuntime.

The benefit of this chance is that plugins (i.e. AnalyticsEngine) and possible ImpromptuServer will return data 2x faster.


jtnelson:
TODO
* Unit test for `LazyTrackingResultDatasets`
* Think through how this affects `ImpromptuServer`. Does it receive an instance of ResultDataset when invoking plugins (I don't think so). So, is there any use of it using this?
* Test that, when a plugin deserializings a dataset it receives from `ConcourseRuntime` that it deserializes to an instanceof LazyTrackingResultDataset with tracking already enabled.

PR #380, opened 2019-07-06
Defining a new thrift exception type, and creating new abstractions.

Tylerwbrown:
Most of the exceptions were just conceptual wrapper around a message, with a lot of duplicate code. There was an exception (ha) to this rule however, so I created an GenericEmptyException as well as GenericMessageException.


jtnelson:
@Tylerwbrown
Exceptions in Thrift must be defined in the service definition (see https://diwakergupta.github.io/thrift-missing-guide/#_types). The duplicated code you refactored is actually auto-generated by Thrift and is necessary for Thrift's RPC and de/serialization protocols to work properly.

There's no advantage to trying to make abstractions or refactor the auto-generated code. 

For this change, you just need to
1. add a new exception type to exceptions.thrift
2. Run the thrift generator for the exceptions.thrift file to create the Java types
3. Add a block in ConcourseServer to translate the relevant Java exception type to the desired Thrift exception type
4. Add a block in ConcourseThriftDriver to translate the relevant Thrift exception type to the desired Java type.

Just basically pattern match how exceptions are already handled. There's no need to refactor.

jtnelson:
This is older, but look at https://github.com/cinchapi/concourse/pull/10 for an example of adding an exception type. The exception has to be defined within the thrift files and each method that can throw the exception must declare it. This mean you'd need to add the new exception type to all the methods defined in `interface/module/calcuate.thrift`

Tylerwbrown:
@jtnelson oh alright, I'll revert those changes.

Should we consider moving our generated code to a different source folder? It would be cool to differentiate between normal, modifiable code from autogenerated code. I see online that some people use `generate-src` as an idiom, as opposed to just putting it all in `src`.

jtnelson:
We could eventually. There’s no harm to the current approach and changing it would require changing things within the build process/system so it’s not something I want to tackle now.

JEFF NELSON
Founder & CEO | Cinchapi

Discover how we make data make sense

Tylerwbrown:
When you referred to `ConcourseServer` did you mean `ClientExceptionTranslator` like we had previously talked about? Or is there a spot in `ConcourseServer` as well?

concourse-server/src/main/java/com/cinchapi/concourse/server/aop/ClientExceptionTranslator.java
```diff
@@ -15,6 +15,8 @@
  */
 package com.cinchapi.concourse.server.aop;
 
+import com.cinchapi.concourse.thrift.*;
```
jtnelson: @Tylerwbrown remove * imports

PR #389, opened 2019-07-28
Feature/con 650

robertcli:
@jtnelson Let me know if there's anything missing. In the video kwam sent you I expressed concern that it felt very quick?

### DIFF
--- a/concourse-driver-java/src/main/java/com/cinchapi/concourse/Point.java
+++ b/concourse-driver-java/src/main/java/com/cinchapi/concourse/Point.java
@@ -20,12 +20,23 @@ import javax.annotation.concurrent.Immutable;
 @Immutable
 public final class Point implements Comparable<Point> {
 
+    public static float MIN_LATITUDE = Float.valueOf("-90.0000");
+    public static float MAX_LATITUDE = Float.valueOf("90.0000");
+    public static float MIN_LONGITUDE = Float.valueOf("-180.0000");
+    public static float MAX_LONGITUDE = Float.valueOf("180.0000");
+
     private float latitude;
     private float longitude;
 
     public Point(float latitude, float longitude) {
-        this.latitude = latitude;
-        this.longitude = longitude;
+        if(isValidLatitude(latitude) && isValidLongitude(longitude)) {
+            this.latitude = latitude;
+            this.longitude = longitude;
+        }
+        else {
+            throw new IllegalArgumentException(
+                    "The specified latitude and longitude contain inproper values, please respect min, max values.");
+        }
     }
 
     public static Point to(float latitude, float longitude) {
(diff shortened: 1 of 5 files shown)


concourse-driver-java/src/main/java/com/cinchapi/concourse/Point.java
```diff
@@ -0,0 +1,69 @@
...
+            throw new IllegalArgumentException(
+                    "The specified x and y contain inproper values, please respect min, max values.");
+        }
+    }
+
+    public static Point to(float x, float y) {
+        return new Point(x, y);
+    }
+
+    public float getY() {
```
jtnelson: just name this method `y()`

concourse-driver-java/src/main/java/com/cinchapi/concourse/Point.java
```diff
@@ -0,0 +1,69 @@
...
+        return y;
+    }
+
+    public float getX() {
```
jtnelson: just name this method `x()`

concourse-driver-java/src/main/java/com/cinchapi/concourse/Point.java
```diff
@@ -0,0 +1,69 @@
...
+ * limitations under the License.
+ */
+package com.cinchapi.concourse;
+
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public final class Point {
+
+    /**
```
jtnelson: There needs to be more javadoc in this class. Please see other classes in the codebase for conventions on how methods, etc javadoced

concourse-driver-java/src/main/java/com/cinchapi/concourse/Point.java
```diff
@@ -0,0 +1,69 @@
...
+     * Valid latitude values are between -90 and 90, both inclusive.
+     */
+    public static float MIN_Y = Float.valueOf("-90.0000");
```
jtnelson: why are there min and max values if a Point is generic as opposed to being lat/long?

interface/shared.thrift
```diff
@@ -48,7 +48,8 @@ enum Operator {
   BETWEEN = 9,
   LINKS_TO = 10,
   LIKE = 11,
-  NOT_LIKE = 12
+  NOT_LIKE = 12,
+  WITHIN = 13
```
jtnelson: remove this, since this PR is not tackling operators for Points

concourse-driver-java/src/main/java/com/cinchapi/concourse/Point.java
```diff
@@ -0,0 +1,69 @@
...
+    public static float MAX_Y = Float.valueOf("90.0000");
+
+    /**
+     * Valid longitude values are between -180 and 180, both inclusive.
+     */
+    public static float MIN_X = Float.valueOf("-180.0000");
+    public static float MAX_X = Float.valueOf("180.0000");
+
+    private float x;
```
jtnelson: these should be doubles instead of floats

PR #391, opened 2019-08-01
[WIP] Java 11 upgrade

Tylerwbrown:
### Notes about the upgrade
- ClassLoader can no longer be cast to URLClassLoader. This was never an intended feature in Java apparently, and was just an implementation detail, which is why although Java 9 is described as "backwards compatible", this was a breaking change. Luckily the only place where we used it, it wasn't necessary.
- We're getting warnings about illegal reflective access now, some important notes:
    - "The first reflective-access operation to any such package causes a warning to be issued, but no warnings are issued after that point." 
    _This means I don't have a good sense of how widespread this issue is in our codebase, but I know we use reflection quite a bit, so it would make sense that this is a large issue_
    - "This warning cannot be suppressed."
    - "All illegal access operations will be denied in a future release"

### TODO:
- Fix remaining test failure (currently only one left)
- Fix Javadocs (currently disabled)


Tylerwbrown:
I don't believe that Javadocs are functional in the current version of Gradle + Java 11.  

The issue is detailed in the following spots:
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=920020 (maven)
- https://bugs.openjdk.java.net/browse/JDK-8212233 (maven)
- https://github.com/gradle/gradle/issues/8267 (gradle)
In the first source it claimed it was fixed in recent versions of openjdk Java-11, but I have the most recent version and it's still not working.
The second claims it was fixed in Java 12 and they had issues backporting it to Java 11, I tried Java 12 and it still didn't build.
The two above sources were both in Maven, and the third is specific to Gradle. This one doesn't seem to have been resolved, which coincides with my testing. 

@jtnelson if you could take a look it'd be appreciated

PR #395, opened 2019-08-08
[WIP] [BLOCKED] CON-206

Tylerwbrown:
Ticket: https://cinchapi.atlassian.net/projects/CON/issues/CON-206?filter=allopenissues&orderby=priority%20DESC

BLOCKED BY: https://github.com/cinchapi/accent4j/pull/10
TODO:
- Add unit tests
- remove the two functions provided by the above PR and use those.


concourse-cli/src/main/java/com/cinchapi/concourse/cli/CommandLineInterface.java
```diff
@@ -44,7 +44,7 @@
  * 
  * @author Jeff Nelson
  */
-public abstract class CommandLineInterface {
+public abstract class CommandLineInterface<ChildOptions extends Options> {
```
jtnelson: While I like this, this is a potentially breaking change.

At the very least, it causes not conforming code to throw a warning. The warning could be easily mitigated by providing the type for ChildOptions, but many third party extensions that leverage this library use a private inner class for options, which cannot be used as a type parameter.

These can be easily fixed, but the larger point is that this change would break existing code and is therefore out of scope for a minor release.

concourse-export/src/main/java/com.cinchapi.concourse.exporter/ExportCli.java
```diff
@@ -0,0 +1,112 @@
...
+
+public class ExportCli extends CommandLineInterface<ExportOptions> {
+    @Override
+    protected void doTask() {
+        final Set<Long> recordIDs = options.records.size() > 0 ? options.records
+                : concourse.inventory();
+
+        Path path = createFile();
+
+        final Iterable<Map<String, Object>> records = (options.criteria != null
```
jtnelson: You could just do concourse.select(records) or concourse.select(ccl) instead of getting key/values one by one

concourse-export/src/main/java/com.cinchapi.concourse.exporter/ExportCli.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    @Override
+    protected ExportOptions getOptions() {
+        return new ExportOptions();
+    }
+
+    private Path createFile() {
+        try {
```
jtnelson: Should probably add parameters to take `order` and `page` for the data to be exported

concourse-export/src/main/java/com.cinchapi.concourse.exporter/ExportCli.java
```diff
@@ -0,0 +1,112 @@
...
+            ConsoleReader reader = new ConsoleReader();
+            System.out
```
jtnelson: Should also add parameter for fields to export for each record

concourse-export/src/main/java/com.cinchapi.concourse.exporter/ExportCli.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.stream.Collectors;
+
+import jline.console.ConsoleReader;
+
+import com.cinchapi.common.base.AnyStrings;
+import com.cinchapi.concourse.cli.CommandLineInterface;
+
+public class ExportCli extends CommandLineInterface<ExportOptions> {
+    @Override
+    protected void doTask() {
```
jtnelson: If no file is specified the CLI should just print the content to sysout

concourse-cli/src/main/java/com/cinchapi/concourse/cli/CommandLineInterface.java
Tylerwbrown: My understanding of this was it's not a breaking change (i.e ImportCLI still works without specifying a generic).
In my mind it's like the difference between using `List` and `List<Type>`. The former works but just lacks the same level of type safety.

Am I missing something? Or are we just trying to mitigate potential warnings that people who conform to this interface get?

jtnelson: warning mitigation

concourse-export/src/main/java/com.cinchapi.concourse.exporter/ExportCli.java
Tylerwbrown: Awesome! I was looking for exactly that type of method but couldn't find it, so I assumed it didn't exist. Thanks!

concourse-export/src/main/java/com.cinchapi.concourse.exporter/ExportCli.java
Tylerwbrown: Should we do something in between the two? I.E currently I recursively call the function until a proper input is given.
I feel like a good middle ground might be asking twice or three times (to account for possible misspelling) or asking up front if they want to print to the console or export to a file.

jtnelson: No. The purpose of having the flags to take in a file is for the CLI to be run non-interactively.

If no file is provided via the flag, you should not prompt the user for input. That breaks non-interactive workflows that want to have the content of the CLI printed to stdout.

Tylerwbrown: Oh, in https://cinchapi.atlassian.net/projects/CON/issues/CON-206?filter=allopenissues&orderby=priority%20DESC file isn't mentioned as a parameter.

So we want to add `file` as well as the other ones you mentioned above? (order/page/fields)?

jtnelson: Yes

concourse-export/src/main/java/com.cinchapi.concourse.exporter/ExportCli.java
Tylerwbrown: What method should I call in the situation that they supply both records and a CCL?
There's no function for both ccl & records. There is one where the types align `(String, Collection<Long>` so I was fooled for a second, but the `String` is a `key`, not a `ccl`.

Tylerwbrown: If there is none I can obviously just stream->filter->collect for that specific case, just figured there might be.

PR #398, opened 2019-08-17
Improve sorting performance

jtnelson:


jtnelson:
TODO
- test gather implementation
- StoreSorter must decided whether to use gather or select based on the number of records that are being pulled and insights about the state of the Engine
- Add ability to sort record ids in the Engine natively?

concourse-server/src/main/java/com/cinchapi/concourse/server/storage/db/SecondaryRecord.java
```diff
@@ -117,6 +119,29 @@
...
+            boolean historical = timestamp == Time.NONE;
+            Set<Value> values = Sets.newLinkedHashSet();
+            for (Value value : history.keySet()) {
+                Set<PrimaryKey> records = historical ? get(value, timestamp)
+                        : get(value);
+                if(records.contains(record)) {
+                    values.add(value);
+                }
+            }
+            return values;
```
Tylerwbrown: `gather` seems like a generic filter operation. Does `Set` and/or `accent4j` not having something that can shorten this?

For example, some idealized Java:

```Java
return history.keys()
    .filter(v -> historical ? get(v, timestamp) : get(v))
```
what I expect vanilla Java can do right now:
```Java
return history.keySet()
    .stream()
    .filter(v -> historical ? get(v, timestamp) : get(v))
    .collect(Collectors.toSet()); // didn't check to see if `toSet` was a method, or `toLinkedHashSet` if we need that specific kind of set for some reason and it's not just noise.
```
as opposed to the original:
```Java
Set<Value> values = Sets.newLinkedHashSet();
for (Value value : history.keySet()) {
    Set<PrimaryKey> records = historical ? get(value, timestamp)
            : get(value);
    if(records.contains(record)) {
        values.add(value);
    }
}
return values;
```

I don't see any advantage in keeping the extra verbosity here.

Tylerwbrown: If Java had short inner method function syntax (like Scala/Kotlin/Python/Ruby/Swift/Haskell/etc.) then the following would be possible, and probably the best, version:
```Java
def Value.records = historical ? get(this, timestamp) : get(this)

return history.keys().filter(v -> v.records.contains(record))
```
granted that's also an extension function, which only Scala (3) and Kotlin support well IMO, although other languages have support too.

You could even pull out that extension function to a scope that makes sense, shortening even more code. It's like a snowball effect.

Java can get extension methods from Manifold and Lombok, but it's experimental in Lombok and you have to pay to use the plugin for Manifold so probably not worth it.

jtnelson: @Tylerwbrown Streams in Java 8 are pretty slow compared to regular for loops. I think later VMs have optimized them, but for low level code we just use for loops so there's no performance regression.

Tylerwbrown: @jtnelson Ah I see. We could still add a filter method to accent4j though right? It obviously wouldn't look as nice as the ideal, but at least it'd abstract away all the repetition for what is just a simple filter. 
```
Sets.filter(history.keySet(), v -> 
    (historical ? get(v, timestamp) : get(v)).contains(record));
```
Get some overhead in terms of code complexity and chaining calls would look a lot worse, but I still think it's worth abstracting since filtering is such a common operation.

jtnelson:
TODO:
- test that gather is faster in Engine for getting values for multiple key/records
- Make StoreSorter lookup use gather

PR #430, opened 2021-08-22
Detect and handle chunks larger than Integer#MAX_VALUE [develop]

jtnelson:


jtnelson:
Test upgrade tasks to ensure they still work as expected since the Manifest not requires more space to accommodate long positions.

PR #440, opened 2021-08-30
Add hidden feature to shortcut logic to potentially speed up verifies for data that is known to not exist

jtnelson:


jtnelson:
Need a cross version test that specifically checks verifies of data that doesn't exist. Also, writes should theoretically be faster if verifies are truly faster

PR #477, opened 2022-05-22
Various optimizations and streamlining to make reads that perform a combination of order, pagination and time travel more efficient

jtnelson:


jtnelson:
This PR includes #476

PR #482, opened 2022-08-13
Support for YAML Config

jtnelson:


jtnelson:
Current Blocker:
Need a way to incrementally source config in startup script since both concourse.prefs and concourse.yaml need to be consulted....

jtnelson:
Go to https://github.com/search?p=2&q=org%3Acinchapi+concourse.prefs&type=Code
and
https://github.com/search?p=2&q=org%3Acinchapi+concourse_client.prefs&type=Code

to see all the places in the code where references to the new yaml files needs to be created

Issue #512, opened 2024-05-01
Add "probe" method to check if client is still connected to the server


jtnelson:
- Add `probe` method to Concourse client
- Add `probe` method to the ConnectionPool that pulls a random connection and tests whether it is still connected
- Implementation in 0.11 can call inventory() to see if the connection is still alive
- In 0.12 add a thrift method called `probe` that returns true
- In the client catch an exception for a broken pipe or inability to write and return false
- Consider adding a `reconnect()` method that can be used to reestablish the connection if the probe fails

```
/**
 * Probes the current connection to the server.
 * <p>
 * This method checks if the client is still connected to the server by sending a query or signal
 * and awaiting a response. It is used to ensure the client's ability to communicate with the server
 * remains intact.
 * 
 * @return boolean - {@code true} if the client is currently connected to the server,
 *                   {@code false} otherwise.
 */
public boolean probe() {
    // implementation code here
}
```

PR #515, opened 2024-05-02
GH-512: Add ping method to facilitate connection healthcheck between client and server

jtnelson:


jtnelson:
What are the semantics of adding a ping method to
- ConnectionPool
- Runway

PR #520, opened 2024-05-18
Reduce Manifest Memory Consumption

jtnelson:


jtnelson:
When the setting is false, the performance and heap usage is worse than the status quo because the start and end positions are not stored as long. We need to figure out a way to have the Range hold longs when the setting is false.

PR #572, opened 2025-05-02
Improve large term index deduplication for container OS systems

jtnelson:
Rename OffHeapTextSet to the more specific LargeTermIndexDeduplicator to make its purpose clearer so it isn't seen as a more general purpose utility. Also, added a brute force implementation to use as a fallback when we can calculate that there isn't enough native memory to even try to use a chronicle set. This should prevent OOMkilled on K8s where native memory isn't elastic


concourse-server/src/main/java/com/cinchapi/concourse/server/storage/db/search/LargeTermIndexDeduplicator.java
```diff
@@ -0,0 +1,813 @@
...
+     * {@code term.
+     * 
+     * @param expectedInsertions
+     * 
+     * @return the {@link LargeTermIndexDeduplicator}
+     */
+    public static LargeTermIndexDeduplicator create(char[] term,
+            int expectedInsertions) {
+        LargeTermIndexDeduplicator deduplicator;
+        long estimatedMemoryRequired = (long) 4
```
jtnelson, line 119: This is too high. The memory requirement is only relevant for Chronicle (brute force doesn't use extra memory and b+ tree uses disk space). With Chronicle, the bytes are going to be written out. And the assumption is that the text is ASCII so we don't need to multiple by 4 here.

